PR #99, opened 2018-03-28
Fix incorrect method of checking path existence

LazarusX:
API document:
* `fs.exists`: https://nodejs.org/api/fs.html#fs_fs_exists_path_callback
* `fse.pathExists`: https://github.com/jprichardson/node-fs-extra/blob/master/docs/pathExists.md


adashen:
Is this a issue we need to publish a hot fix for it?

PR #105, opened 2018-04-12
Add support for buildOptions

adashen:
Add "buildOptions" into module.json to customize the build options

### DIFF
--- a/src/container/containerManager.ts
+++ b/src/container/containerManager.ts
@@ -29,7 +29,9 @@ export class ContainerManager {
                 const directory = path.dirname(moduleConfigFilePath);
                 const dockerfilePath = path.join(directory, platforms[platform]);
                 const imageName = Utility.getImage(moduleConfig.image.repository, moduleConfig.image.tag.version, platform);
-                const buildCommand = this.constructBuildCmd(dockerfilePath, imageName, directory);
+                const options = moduleConfig.image.buildOptions;
+                const optionArray: string[] = options && options instanceof Array ? options : undefined;
+                const buildCommand = this.constructBuildCmd(dockerfilePath, imageName, directory, optionArray);
                 if (pushImage) {
                     const pushCommand = this.constructPushCmd(imageName);
                     Executor.runInTerminal(Utility.combineCommands([buildCommand, pushCommand]));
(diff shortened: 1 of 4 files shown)


src/container/containerManager.ts
```diff
@@ -128,8 +137,20 @@ export class ContainerManager {
...
+    private constructBuildCmd(dockerfilePath: string, imageName: string, contextDir: string, extraOptions ?: string[]): string {
+        let optionString: string = "";
+        if (extraOptions !== undefined) {
+            const filteredOption = extraOptions.filter((value, index) => {
+                const trimmed = value.trim();
+                return ! (trimmed.startsWith("--rm") ||
+                            trimmed.startsWith("--tag") ||
+                            trimmed.startsWith("-t") ||
+                            trimmed.startsWith("--file") ||
+                            trimmed.startsWith("-f"));
```
LazarusX: This will also filter options like `--target` and `--force-rm`, right?

LazarusX: Sorry. I am wrong. These start with double `-`

LazarusX: But I still believe we should better check the boundary of the options. Say, in the future, a new option `--file-balabala` is added, it will be filtered.

formulahendry: Agree, We could add a space (e.g. `trimmed.startsWith("--file ")`) to check

LazarusX: I guess simply adding a space may not be enough, because the option could have default values. For example, `--rm` has the same effect as `--rm true`.

test/utility.test.ts
```diff
@@ -48,7 +48,20 @@ suite("utility tests", () => {
     assert.equal(moduleToImageMap.size, 4);
     assert.equal(imageToDockerfileMap.size, 4);
     assert.equal(imageToBuildOptions.size, 4);
-    assert.equal(imageToBuildOptions.get("localhost:5000/samplemodule:0.0.1-amd64").length, 7);
+    assert.equal(imageToBuildOptions.get("localhost:5000/samplemodule:0.0.1-amd64").length, 8);
+
+    const filteredOption = imageToBuildOptions.get("localhost:5000/samplemodule:0.0.1-amd64").filter((value, index) => {
```
formulahendry: We'd better to wrap the filter logic into a method, so that the test code and prod code could share.

adashen: The current logic is wrapped in the private method constructBuildCmd.  I am thinking of add anther utility public method, but seems a little bit over-wrapped. Another engineer task could in our backlog is to wrap the "module.json" schema into a class. So all the stuff could be checked in the class. I will raise an issue in github for tracking.

src/container/containerManager.ts
```diff
@@ -142,11 +142,13 @@ export class ContainerManager {
...
             const filteredOption = extraOptions.filter((value, index) => {
                 const trimmed = value.trim();
-                return ! (trimmed.startsWith("--rm") ||
-                            trimmed.startsWith("--tag") ||
-                            trimmed.startsWith("-t") ||
-                            trimmed.startsWith("--file") ||
-                            trimmed.startsWith("-f"));
+                const parsedOption: string[] = trimmed.split(/\s+/g);
+                return parsedOption.length > 0 &&
+                       ! ((parsedOption[0] === "--rm" ) ||
```
formulahendry: Better to use `!["--rm", "--tag" ...].includes(parsedOption[0])`. It is simpler.

PR #107, opened 2018-04-13
Add more IntelliSense features

LazarusX:
Add more IntelliSense features:
- Image URL validation
- Dockerfile path validation
- Hover preview of Dockerfile content on image URL
- Go-to-Dockerfile on image URL


src/extension.ts
```diff
@@ -21,9 +21,22 @@ export function activate(context: vscode.ExtensionContext) {
 
     Utility.registerDebugTelemetryListener();
 
-    context.subscriptions.push(vscode.languages.registerCompletionItemProvider([{ language: "json" }, { language: "jsonc" }], new JsonCompletionItemProvider(), "\"", ".", ":"));
-
-    const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel(Constants.edgeChannel);
+    const configIntelliSenseProvider = new ConfigIntelliSenseProvider();
```
adashen: Is there any reason we put three kinds of provider into one class?

src/common/configIntelliSenseProvider.ts
```diff
@@ -0,0 +1,336 @@
...
+        if (this.locationMatch(location, Constants.imgDeploymentManifestJsonPath)) {
+            const moduleToImageMap: Map<string, string> = new Map();
+            const imageToDockerfileMap: Map<string, string> = new Map();
+
+            try {
+                await Utility.setSlnModulesMap(path.dirname(document.uri.fsPath), moduleToImageMap, imageToDockerfileMap);
+
+                const node: parser.Node = location.previousNode;
+                const imageUrl: string = node.value;
+                const imageUrlUnwrapped: string = imageUrl.slice(2, -1); // remove the wrapping "${" and "}"
```
adashen: edge case. If the imageUrl is not start with ${ and end with }, And after slicing imageUrl.slice(2, -1), the value could be got from the map.

src/common/configIntelliSenseProvider.ts
```diff
@@ -0,0 +1,336 @@
...
+            } catch {
+                return undefined;
+            }
+        }
+
+        return undefined;
+    }
+
+    public async provideDefinition(document: vscode.TextDocument, position: vscode.Position): Promise<vscode.Definition> {
+        const location: parser.Location = parser.getLocation(document.getText(), document.offsetAt(position));
```
adashen: duplicate code between provideDefinition and provideHover, extract an utility instead?

src/common/configIntelliSenseProvider.ts
```diff
@@ -0,0 +1,336 @@
...
+                const moduleName: string = moduleNode.children[0].value; // the node property name is stored in its first child node
+                const imgJsonPath: string[] = Constants.imgDeploymentManifestJsonPath.slice(Constants.moduleNameDeploymentManifestJsonPathIndex + 1); // image node JSON path relative to module node
+                const imageNode: parser.Node = parser.findNodeAtLocation(moduleNode.children[1], imgJsonPath); // the node value is stored in its second child node
+
+                if (!imageNode) {
+                    continue;
+                }
+
+                const imageUrl: string = imageNode.value;
+                if (imageUrl.search(pattern) !== -1) {
```
adashen: !== -1 or ===0?

src/extension.ts
LazarusX: I have split it into corresponding classes.

src/common/configIntelliSenseProvider.ts
LazarusX: Fixed. Thanks.

src/common/configIntelliSenseProvider.ts
LazarusX: Fixed. Thanks.

src/common/configIntelliSenseProvider.ts
LazarusX: Fixed. Thanks.

PR #141, opened 2018-05-07
Respect .env file

formulahendry:


src/container/containerManager.ts
```diff
@@ -22,11 +22,12 @@ export class ContainerManager {
         const moduleConfigFilePath: string = await Utility.getInputFilePath(fileUri, Constants.moduleConfigFileNamePattern, Constants.moduleConfigFile, `${event}.selectModuleConfigFile`);
 
         if (moduleConfigFilePath) {
+            const directory = path.dirname(moduleConfigFilePath);
```
adashen: catch exception of loading env? fail loading env should not fail the build task

src/container/containerManager.ts
```diff
@@ -73,6 +74,7 @@ export class ContainerManager {
         const imageToDockerfileMap: Map<string, string> = new Map();
         const imageToBuildOptions: Map<string, string[]> = new Map();
         const slnPath: string = path.dirname(templateFile);
+        await Utility.loadEnv(path.join(slnPath, Constants.envFile));
```
adashen: same as above

test/extension.test.ts
```diff
@@ -10,13 +10,6 @@ suite("Extension Tests", () => {
         assert.ok(vscode.extensions.getExtension(Constants.ExtensionId));
     });
 
-    test("should activate", function() {
```
formulahendry: Extension could only be activated when extensionDependencies are installed, so remove it in CI.

PR #142, opened 2018-05-08
Automatically start local registry

### LINKED ISSUE #129: Automatically start local registry if detect the user is using local docker registry on devbox
@jonbgallant mentioned:
For example, when I start the local registry, I print messages to the screen letting the user know what it is doing.
Here’s the iotedgedev code that inits the local registry
https://github.com/Azure/iotedgedev/blob/master/iotedgedev/dockercls.py#L38
It would be great if the extension could do something similar.
Basically, the code is “if localhost, start local registry server”
That would allow the extension with all the default settings presented in the extension.
Alternatively, you could recognize that they are using localhost and let them know exactly how to start it.

formulahendry:
Resolve #129

### DIFF
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -7,7 +7,7 @@ import * as fse from "fs-extra";
 import * as os from "os";
 import * as path from "path";
 import * as vscode from "vscode";
-import { Constants } from "./constants";
+import { Constants, ContainerState } from "./constants";
 import { Executor } from "./executor";
 import { TelemetryClient } from "./telemetryClient";
 import { UserCancelledError } from "./UserCancelledError";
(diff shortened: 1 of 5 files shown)


src/common/utility.ts
```diff
@@ -293,4 +293,36 @@ export class Utility {
             }
         } catch (error) {}
     }
+
+    public static initLocalRegistry(images: string[]) {
+        try {
+            let port;
+            for (const image of images) {
+                const matches = /^localhost:(\d+)\/.+$/.exec(image);
```
adashen: only "/" ?

formulahendry: Yes.

PR #144, opened 2018-05-10
Add option to import images from ACR

LazarusX:
Also in this PR:
* Remove unused imports
* Resolve security vulnerability
* Update extension category information per VSCode 1.23.0's change


src/edge/edgeManager.ts
```diff
@@ -54,14 +52,14 @@ export class EdgeManager {
         mapObj.set(Constants.moduleImagePlaceholder, imageName);
         mapObj.set(Constants.moduleFolderPlaceholder, moduleName);
         const deploymentGenerated: string = Utility.replaceAll(data, mapObj);
-        await fse.writeFile(targetDeployment, deploymentGenerated, {encoding: "utf8"});
+        await fse.writeFile(targetDeployment, deploymentGenerated, { encoding: "utf8" });
```
LazarusX: Sorry for this kind of distractive change. This is the side-effect of a "format code"

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+                        progress.report({ message: `Enabling admin user for registry "${registryUrl}"...` });
+                        await client.registries.update(resourceGroup, registryName, { adminUserEnabled: true });
+                    }
+
+                    progress.report({ message: `Fetching admin user credentials for registry "${registryUrl}"...` });
+                    const creds: RegistryListCredentialsResult = await client.registries.listCredentials(resourceGroup, registryName);
+                    const username: string = creds.username;
+                    const password: string = creds.passwords[0].value;
+
+                    Executor.runInTerminal(`docker login -u ${username} -p ${password} ${registryUrl}`);
```
LazarusX: I will update in the future to provide password to docker login via STDIN. https://github.com/Microsoft/vscode-azure-iot-edge/issues/145

package.json
```diff
@@ -201,28 +201,35 @@
...
     "fs-extra": "^4.0.2",
     "is-port-reachable": "^2.0.0",
     "jsonc-parser": "^1.0.1",
+    "request": "^2.85.0",
+    "request-promise": "^4.2.2",
     "strip-json-comments": "^2.0.1",
     "vscode-extension-telemetry": "0.0.8"
   },
   "extensionDependencies": [
+    "ms-vscode.azure-account",
```
adashen: azure account is already a dependency of azure-iot-toolkit. Depends on the parsing strategy of vscode, and we need keep an eye of the installation success rate

src/common/constants.ts
```diff
@@ -26,11 +26,13 @@ export class Constants {
     public static LANGUAGE_NODE = "Node.js Module";
     public static LANGUAGE_PYTHON = "Python Module";
     public static CSHARP_FUNCTION = "Azure Functions - C#";
+    public static ACR_MODULE = "Existing Module on Azure Container Registry";
```
adashen: M lower case?

package.json
LazarusX: Got it. I just want to make it more explicit as Edge extension now directly depends on azure-account.

src/common/constants.ts
LazarusX: I use title case to align with other options.

src/common/models/IPartialList.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+"use strict";
+
+export interface IPartialList<T> extends Array<T> {
```
adashen: is this interface defined in azure-arm-containerregistry/lib/models

src/common/models/IAzureClient.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+"use strict";
+import { IPartialList } from "./IPartialList";
+
+export interface IAzureClient<T> {
```
adashen: is there similar interface defined in azure-arm-containerregistry/lib/models?

src/common/utility.ts
```diff
@@ -292,7 +294,16 @@ export class Utility {
...
                 }
             }
-        } catch (error) {}
+        } catch (error) { }
+    }
+
+    // The Azure API of listing resources is paginated. This method will follow the links and return all resources
+    public static async listAllAzureResource<T>(client: IAzureClient<T>, first: Promise<IPartialList<T>>): Promise<T[]> {
+        const all: T[] = [];
+        for (let list = await first; list.length || list.nextLink; list = list.nextLink ? await client.listNext(list.nextLink) : []) {
```
adashen: change the condition to list !== undefined ?

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        this.loginToAcr(acrRegistryItem.registry, acrRegistryItem.azureSubscription);
+
+        return acrTagItem.description;
+    }
+
+    private async selectAcrRegistry(): Promise<AcrRegistryQuickPickItem> {
+        if (!(await this.azureAccount.waitForLogin())) {
+            await vscode.commands.executeCommand("azure-account.askForLogin");
```
formulahendry: What if user cancel for "azure-account.askForLogin"?  You could refer to https://github.com/Microsoft/vscode-azure-iot-toolkit/blob/ed583eaebcd336a94ed2a97e740c5302cd6f2b4a/src/iotHubResourceExplorer.ts#L173

src/common/models/IAzureClient.ts
LazarusX: Good point. Actually these snippets are based VSCode Azure Account SDK's sample code which is service-agnostic. I will update it to be more specific to our scenarios and thus more concise. Thanks.

src/edge/edgeManager.ts
```diff
@@ -341,6 +342,10 @@ export class EdgeManager {
                 label: Constants.CSHARP_FUNCTION,
                 description: Constants.CSHARP_FUNCTION_DESCRIPTION,
             },
+            {
+                label: Constants.ACR_MODULE,
```
adashen: instead of have ACR_MODULE on the same level with EXISTING_MODULE, how about make it sub of EXISTING_MODULE, user could either input or select form ACR?

src/common/utility.ts
LazarusX: Updated. Thanks.

src/edge/acrManager.ts
LazarusX: Fixed. Thanks.

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+import * as vscode from "vscode";
+import { Executor } from "../common/executor";
+import { UserCancelledError } from "../common/UserCancelledError";
+import { Utility } from "../common/utility";
+import { AcrRegistryQuickPickItem } from "../container/models/AcrRegistryQuickPickItem";
+import { AzureAccount, AzureSession, AzureSubscription } from "../typings/azure-account.api";
+
+export class AcrManager {
+    private readonly azureAccount: AzureAccount;
+    private refreshTokenArc: string;
```
adashen: Arc or Acr?

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+            await this.azureAccount.waitForFilters();
+            const registryPromises: Array<Promise<AcrRegistryQuickPickItem[]>> = [];
+            for (const azureSubscription of this.azureAccount.filters) {
+                const client = new ContainerRegistryManagementClient(
+                    azureSubscription.session.credentials,
+                    azureSubscription.subscription.subscriptionId!,
+                );
+
+                registryPromises.push(
+                    Utility.listAllAzureResource(client.registries, client.registries.list())
```
adashen: use await Utility.listAllAzureResource, instead of then? to make code more easily read

src/edge/acrManager.ts
LazarusX: This name is based on vscode-docker's pattern (for example https://github.com/Microsoft/vscode-docker/blob/master/explorer/models/azureRegistryNodes.ts#L52-L53), as I cannot find ACR's data plane REST API document.

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    private async selectAcrRepo(registryUrl: string, session: AzureSession): Promise<vscode.QuickPickItem> {
+        const acrRepoItem: vscode.QuickPickItem = await vscode.window.showQuickPick(this.loadAcrRepoItems(registryUrl, session), { placeHolder: "Select Repository", ignoreFocusOut: true });
+        return acrRepoItem;
+    }
+
+    private async loadAcrRepoItems(registryUrl: string, session: AzureSession): Promise<vscode.QuickPickItem[]> {
+        try {
+            const { accessToken, refreshToken } = await this.acquireToken(session);
+            this.refreshTokenArc = await this.acquireRefreshTokenArc(registryUrl, session.tenantId, refreshToken, accessToken);
```
adashen: acr accesstoken will not in this response?

adashen: Can we just use acquireAccessTokenAcr with the access token and refresh token? Besides, shall we cache the refresh token?

src/edge/acrManager.ts
LazarusX: Adding `await` here will suspend the execution and thus affects the performance.

src/edge/edgeManager.ts
LazarusX: I am afraid existing UI controls provided by VS Code don't allow the coexistence of an input box and a quick pick item.

src/edge/acrManager.ts
```diff
@@ -0,0 +1,224 @@
...
+            const loginMessage = adminUserEnabled ? `Looks like the registry "${registryUrl}" has admin user enabled. `
+                + `Would you like to run "docker login" with the admin user credentials?`
+                : `Looks like the registry "${registryUrl}" does not have admin user enabled. `
+                + `Would you like to enable admin user and run "docker login" with the admin user credentials?`;
+            const option: string = await vscode.window.showInformationMessage(loginMessage, "Yes", "No");
+            if (option === "Yes") {
+                vscode.window.withProgress({ title: `Logging in to registry "${registryUrl}"...`, location: vscode.ProgressLocation.Window }, async (progress) => {
+                    const registryName: string = registry.name;
+                    const resourceGroup: string = registry.id.slice(registry.id.toLowerCase().search("resourcegroups/") + "resourcegroups/".length, registry.id.toLowerCase().search("/providers/"));
+                    const client = new ContainerRegistryManagementClient(
```
adashen: the ContainerRegistryManagementClient should only need one instance after login right?

src/edge/acrManager.ts
LazarusX: No, not in the response.

LazarusX: No. The access tokens for listing repos and tags are different in scope (compare Line 91 and Line 164). `refreshTokenArc` is already cached in the object property.

src/edge/acrManager.ts
LazarusX: Fixed. Thanks.

LazarusX: Sorry, but I reverted this change. `ContainerRegistryManagementClient` is on a subscription basis. Since a user may have multiple subscriptions, there are multiple instance created, and when `ContainerRegistryManagementClient` is [firstly introduced]( https://github.com/Microsoft/vscode-azure-iot-edge/pull/144/files#diff-e8e674adc42104c063931eb583165d99R64), we don't know which one he will pick yet.

src/container/acrManager.ts
```diff
@@ -0,0 +1,237 @@
...
+export class AcrManager {
+    private readonly azureAccount: AzureAccount;
+    private acrRefreshToken: string;
+
+    constructor() {
+        this.azureAccount = vscode.extensions.getExtension<AzureAccount>("ms-vscode.azure-account")!.exports;
+    }
+
+    public async selectAcrImage(): Promise<string> {
+        const acrRegistryItem: AcrRegistryQuickPickItem = await this.selectAcrRegistry();
```
adashen: suggest to get the ContainerRegistryManagementClient first in this method. Thus the client could be reused in the flow of this method.

PR #148, opened 2018-05-11
Only respect .env file when current workspace is solution folder and …

formulahendry:
…contains .env file in the root folder

### DIFF
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -284,9 +284,16 @@ export class Utility {
         return undefined;
     }
 
-    public static async loadEnv(envFilePath: string) {
+    public static async loadEnv(filePath: string) {
         try {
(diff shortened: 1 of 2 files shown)


src/common/utility.ts
```diff
@@ -284,9 +284,16 @@ export class Utility {
         return undefined;
     }
 
-    public static async loadEnv(envFilePath: string) {
+    public static async loadEnv(filePath: string) {
         try {
-            if (await fse.pathExists(envFilePath)) {
+            const workspaceFolder = vscode.workspace.getWorkspaceFolder(vscode.Uri.file(filePath));
```
adashen: check multi-root nested scenario. And how about use the vscode.uri selected by user from ui to get the workspacefolder?

PR #150, opened 2018-05-15
Add ACR credentials in IoT Edge runtime instead of docker login

LazarusX:
Per my test, you have to add the ACR credentials in IoT Edge runtime so that the Edge runtime can pull the images; `docker login` does not work.

### DIFF
--- a/src/container/acrManager.ts
+++ b/src/container/acrManager.ts
@@ -203,12 +203,12 @@ export class AcrManager {
             const adminUserEnabled: boolean = registry.adminUserEnabled;
             const registryUrl: string = registry.loginServer;
             const loginMessage = adminUserEnabled ? `Looks like the registry "${registryUrl}" has admin user enabled. `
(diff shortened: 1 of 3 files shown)


src/container/acrManager.ts
```diff
@@ -226,7 +226,7 @@ export class AcrManager {
                     const username: string = creds.username;
                     const password: string = creds.passwords[0].value;
 
-                    Executor.runInTerminal(`docker login -u ${username} -p ${password} ${registryUrl}`);
+                    Executor.runInTerminal(`iotedgectl login --address ${registryUrl} --username ${username} --password ${password}`);
```
formulahendry: On Linux/macOS, it needs to run as sudo

formulahendry: Another concern is that currently all `iotedgectl` logic in controlled in IoT Toolkit extension. After this change, IoT Edge extension would also depend on iotedgectl. When iotedgectl is replace with edgeLet, we need to update both two extensions.

PR #158, opened 2018-05-23
Align files generated by ConvertModule with new C# templates releases

### LINKED ISSUE #157: Files generated by ConvertModule are not consistent with latest version of C# templates
Reported by @czgtest

LazarusX:
Fix #157


src/common/utility.ts
```diff
@@ -143,6 +143,14 @@ export class Utility {
         return null;
     }
 
+    public static async copyTemplateFile(srcPath: string, srcFile: string, targetPath: string, mapObj: Map<string, string>) {
```
adashen: better to name "srcFile" as filename, since the file name is shared in both source and target

PR #198, opened 2018-07-13
Add Stream Analytics Module support

SLdragon:


src/container/streamAnalyticsManager.ts
```diff
@@ -0,0 +1,108 @@
...
+    public getImageName(): string {
+        return this.STREAM_ANALYTICS_IMAGE;
+    }
+
+    public async getJobInfo(streamingJob: StreamAnalyticsPicItem): Promise<object> {
+        const subscriptionId: string = streamingJob.azureSubscription.subscription.id;
+        const id: string = streamingJob.job.id;
+        const resourceGroup: string = id.slice(id.toLowerCase().search("resourcegroups/") + "resourcegroups/".length, id.toLowerCase().search("/providers/"));
+        const jobName: string = streamingJob.job.name;
+        const publishJobUrl: string = `https://management.azure.com${subscriptionId}/resourceGroups/${resourceGroup}` +
```
LazarusX: Should there be a `/` between `management.azure.com` and `${subscriptionId}`?

src/container/streamAnalyticsManager.ts
```diff
@@ -0,0 +1,108 @@
...
+import StreamAnalyticsManagementClient = require("azure-arm-streamanalytics");
+import * as request from "request-promise";
+import * as vscode from "vscode";
+import { UserCancelledError } from "../common/UserCancelledError";
+import { StreamAnalyticsPicItem } from "../container/models/streamAnalyticsPickItem";
+import { AzureAccount, AzureSession } from "../typings/azure-account.api";
+
+export class StreamAnalyticsManager {
+    private readonly azureAccount: AzureAccount;
+    private readonly STREAM_ANALYTICS_IMAGE: string = "microsoft/azureiotedge-azure-stream-analytics:1.0.0-preview006";
```
LazarusX: It seems 1.0.0-preview007 is the latest tag: https://hub.docker.com/r/microsoft/azureiotedge-azure-stream-analytics/tags/

src/container/streamAnalyticsManager.ts
```diff
@@ -0,0 +1,108 @@
...
+                });
+            }
+            return items;
+        } catch (error) {
+            error.message = `Error fetching streaming jobs list: ${error.message}`;
+            throw error;
+        }
+    }
+
+    private async acquireAadToken(session: AzureSession): Promise<{ aadAccessToken: string, aadRefreshToken: string }> {
```
LazarusX: This method looks duplicate with https://github.com/Microsoft/vscode-azure-iot-edge/blob/master/src/container/acrManager.ts#L159-L174.

Can you extract it and make ASA manager and ACR manager share it? I think https://github.com/Microsoft/vscode-azure-iot-edge/blob/master/src/common/utility.ts may be a candiate place to put common methods like this,

src/container/streamAnalyticsManager.ts
```diff
@@ -0,0 +1,108 @@
...
+
+        const operationResultUrl = publishResponse.headers.location;
+
+        const jobInfoResult = await request.get(operationResultUrl, {
+            auth: {
+                bearer: aadAccessToken,
+            },
+        });
+
+        try {
```
LazarusX: We might need to wrap above code in `getJobInfo` method with `try/catch` as well in case something wrong happens.

src/container/streamAnalyticsManager.ts
```diff
@@ -0,0 +1,108 @@
...
+            const info = JSON.parse(JSON.parse(jobInfoResult).manifest);
+            return info;
+        } catch (e) {
+            throw new Error("Cannot parse Stream Analytics publish job information!") + e.toString();
```
LazarusX: You are adding an `Error` object with a `string`?

Also a minor comment, use `.` instead of `!` because `.` is a more gentle (just a personal reference anyway).

src/edge/edgeManager.ts
```diff
@@ -345,21 +357,28 @@ export class EdgeManager {
             null, dftValue);
     }
 
-    private async inputImage(module: string, template: string): Promise<{ repositoryName: string, imageName: string }> {
+    private async inputImage(module: string, template: string): Promise<{ repositoryName: string, imageName: string, moduleTwin: object }> {
         let repositoryName: string = "";
         let imageName: string = "";
+        let moduleTwin: object = null;
```
LazarusX: Minor comment: use undefined instead of null: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

src/edge/edgeManager.ts
```diff
@@ -95,7 +102,12 @@ export class EdgeManager {
         const routes = templateJson.moduleContent.$edgeHub["properties.desired"].routes;
         const runtimeSettings = templateJson.moduleContent.$edgeAgent["properties.desired"].runtime.settings;
         const moduleName: string = Utility.getValidModuleName(await this.inputModuleName(targetModulePath, Object.keys(modules)));
-        const { repositoryName, imageName } = await this.inputImage(moduleName, template);
+        const { repositoryName, imageName, moduleTwin } = await this.inputImage(moduleName, template);
+
+        if (template === Constants.STREAM_ANALYTICS && moduleTwin !== null) {
```
LazarusX: Same `null` vs. `undefined`.

src/container/streamAnalyticsManager.ts
formulahendry: Do we need to specify the tag? Could we just use `latest` tag?

SLdragon: I should remove this image URL, because we can retrieve it from publish rest API

src/container/streamAnalyticsManager.ts
SLdragon: Thanks, I have already put it into the utility.ts

src/container/streamAnalyticsManager.ts
SLdragon: Fixed

src/container/streamAnalyticsManager.ts
SLdragon: Fixed

src/container/streamAnalyticsManager.ts
```diff
@@ -29,40 +29,35 @@ export class StreamAnalyticsManager {
...
+            const publishResponse = await request.post(publishJobUrl, {
+                auth: {
+                    bearer: aadAccessToken,
+                },
+                resolveWithFullResponse: true,
+            });
+
+            const operationResultUrl = publishResponse.headers.location;
+
+            await this.sleep(1000);
```
LazarusX: Why do we need to sleep here?

src/container/streamAnalyticsManager.ts
LazarusX: I suggest adding a comment here to give a sample value of `subscriptionId`.

src/container/streamAnalyticsManager.ts
SLdragon: Because the rest api is asynchronous, we need to wait in case of error

LazarusX: Is there any way we can know whether the operation succeeds instead of hard-coding a sleep time?

src/edge/edgeManager.ts
```diff
@@ -483,6 +511,10 @@ export class EdgeManager {
                 label: Constants.EXISTING_MODULE,
                 description: Constants.EXISTING_MODULE_DESCRIPTION,
             },
+            {
+                label: Constants.STREAM_ANALYTICS,
```
LazarusX: I had a offline sync with Xin. Can you rearrange the order to put ASA module below C# Function module? So the new order will be:
* C# Module
* Node.js Module
* Python Module
* Azure Functions - C#
* Azure Stream Analytics
* Existing Module on ACR
* Existing Module

src/edge/edgeManager.ts
```diff
@@ -483,6 +511,10 @@ export class EdgeManager {
                 label: Constants.EXISTING_MODULE,
                 description: Constants.EXISTING_MODULE_DESCRIPTION,
             },
+            {
```
LazarusX: I had a offline sync with Xin. Can you rearrange the order to put ASA module below C# Function module? So the new order will be:
* C# Module
* Node.js Module
* Python Module
* Azure Functions - C#
* Azure Stream Analytics
* Existing Module on ACR
* Existing Module

SLdragon: fixed

src/edge/edgeManager.ts
```diff
@@ -383,11 +386,12 @@ export class EdgeManager {
             const JobInfo: any = await saManager.getJobInfo(job);
             imageName = JobInfo.settings.image;
             moduleTwin = JobInfo.twin.content;
+            createOptions = JobInfo.settings.createOptions ? JSON.stringify(JobInfo.settings.createOptions) : "";
```
LazarusX: I think we should put the string replacement logic here. The outer code can use the returned `createOptions` as a string without doing escaping/unescaping each time.

PR #206, opened 2018-07-27
Add build, build/push, build/run, run commands

LazarusX:


src/container/containerManager.ts
```diff
@@ -107,8 +99,18 @@ export class ContainerManager {
...
             commands.push(this.constructBuildCmd(buildSetting.dockerFile, image, context, buildSetting.buildOption));
-            commands.push(this.constructPushCmd(image));
+            if (push) {
+                commands.push(this.constructPushCmd(image));
+            }
         });
+
+        if (run) {
+            // A temporary hack to keep the command running in a dedicated terminal
+            commands.push(`iotedgehubdev start -d "${deployFile}" -v`);
```
adashen: the command script here is same as the one in runSolution. Better to extract one method.

PR #208, opened 2018-07-30
Merge develop branch into master

LazarusX:


package.json
```diff
@@ -100,6 +130,26 @@
...
+        "category": "Azure IoT Edge"
+      },
+      {
+        "command": "azure-iot-edge.runSolution",
+        "title": "Run IoT Edge Solution",
+        "category": "Azure IoT Edge"
+      },
+      {
+        "command": "azure-iot-edge.stopSolution",
+        "title": "Stop IoT Edge Solution",
```
adashen: just come into mind, Actually it will stop edge runtime both for single module scenario and solution mode right?
 So how about rename the title: Stop IoT Edge Simulator?

PR #211, opened 2018-07-30
Solve moduleContent issue

adashen:
1. moduleContent issue
2. Refactor code to extract "addmodule" logic
3. Expose 5671 in deployement template


src/container/containerManager.ts
```diff
@@ -113,27 +113,27 @@ export class ContainerManager {
 
     private async generateDeploymentString(templateFile: string,
```
LazarusX: `generateDeploymentString` now does not generate a string anymore. Suggest updating the name to reflect this.

PR #227, opened 2018-08-15
add support for docker build context path

adashen:
1. now user could specify the docker build context path at "module.json" at "image/contextPath". The path value could be absolute or relative to the current "module.json"
2. Also refactored the code to define the class "buildsettings"
3. @LazarusX some change in intellisense, please have a check

### DIFF
--- a/src/common/buildSettings.ts
+++ b/src/common/buildSettings.ts
@@ -1,0 +1,11 @@
+export class BuildSettings {
+    public readonly options?: string[];
+    public readonly contextPath: string;
+    public readonly dockerFile: string;
+
+    constructor(dockerFile: string, contextPath: string, options?: string[]) {
+        this.dockerFile = dockerFile;
+        this.contextPath = contextPath;
+        this.options = options;
+    }
+}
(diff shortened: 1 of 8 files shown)


src/common/utility.ts
```diff
@@ -267,32 +267,41 @@ export class Utility {
...
         );
     }
 
+    public static getBuildSettings(
+        modulePath: string,
+        dockerFilePath: string,
+        buildOptions?: string[],
+        contextPath?: string): BuildSettings {
+        const optionArray = (buildOptions && buildOptions instanceof Array) ? buildOptions : undefined;
+        const context = contextPath ? path.join(modulePath, contextPath) : path.dirname(dockerFilePath);
```
LazarusX: I think `path.resolve` is better than `path.join` here as `path.resolve` can handle the case where `contextPath` is absolute.

adashen: good catch, will update

LazarusX:
I have verified the IntelliSense features. Looks good to me.

src/common/buildSettings.ts
```diff
@@ -0,0 +1,11 @@
+export class BuildSettings {
+    public readonly options?: string[];
+    public readonly contextPath: string;
+    public readonly dockerFile: string;
+
+    constructor(dockerFile: string, contextPath: string, options?: string[]) {
```
formulahendry: Could simplify with `constructor(public readonly dockerFile: string, public readonly contextPath: string, public readonly options?: string[])`

PR #272, opened 2018-10-30
Extend CreateOptions

adashen:
Two changes in this changelist
1. Support using JSON object to set createOptions in deployment.template.json. And the default template/addmodule now use JSON object to set createOptions
2. Generate deployment.json will stringfy the createOptions value and chunck it into maximum 8 blocks. from createOptions to createOptions01.....createOptions07

### DIFF
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -485,7 +485,7 @@ export class Utility {
         const re = new RegExp(`.{1,${Constants.TwinValueMaxSize}}`, "g");
         const options = optionStr.match(re);
         if (options.length > Constants.TwinValueMaxChunks) {
-            throw new Error("Size of createOptions too big.");
+            throw new Error("Size of createOptions too big. The maxium size of createOptions is 4K");
         }
         options.map((value, index) => {
             if (index === 0) {


src/common/utility.ts
```diff
@@ -459,6 +475,42 @@ export class Utility {
...
+        let optionStr: string;
+        if (typeof createOptions === "string") {
+            optionStr = createOptions;
+        } else {
+            optionStr = JSON.stringify(createOptions);
+        }
+        const re = new RegExp(`.{1,${Constants.TwinValueMaxSize}}`, "g");
+        const options = optionStr.match(re);
+        if (options.length > Constants.TwinValueMaxChunks) {
+            throw new Error("Size of createOptions too big.");
```
LazarusX: We might need to tell the length limit, i.e., 512.

src/common/utility.ts
```diff
@@ -459,6 +475,42 @@ export class Utility {
...
+
+        return settings;
+    }
+
+    private static serializeCreateOptionsForEachModule(modules: any): any {
+        for (const key in modules) {
+            if (modules.hasOwnProperty(key)) {
+                const moduleVar = modules[key];
+                if (moduleVar.settings && moduleVar.settings.createOptions) {
+                    moduleVar.settings = Utility.serializeCreateOptions(moduleVar.settings, moduleVar.settings.createOptions);
```
LazarusX: I think we also need to support `createOptionsXX` in deployment.template.json.

adashen: It's supported

PR #275, opened 2018-11-02
Handle createOptions with line breaks and spaces

fbeltrao:
After the change to limit the createOptions length I could not build and push my IoT Edge solution.

The problem origin was the fact that the createOptions was copied from Azure Portal, containing line breaks and spaces. The extension wrongly assumed that it was longer than 4K.

The createOptions that was failing was:
```
"{\r\n  \"ExposedPorts\": {\r\n    \"1680/udp\": {}\r\n  },\r\n  \"HostConfig\": {\r\n    \"PortBindings\": {\r\n      \"1680/udp\": [\r\n        {\r\n          \"HostPort\": \"1680\",\r\n          \"HostIp\":\"172.17.0.1\"\r\n        }\r\n      ]\r\n    }\r\n  }\r\n}\r\n\r\n"
```

The proposed change will minify the createOptions string using `JSON.stringify(JSON.parse(createOptions), null, 0)`.

Additionally, the error message was improved:
- module where the error happened is displayed
- typo was fixed.

Please let me know if you need more info/changes

Cheers

### DIFF
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -15,7 +15,6 @@ import { Constants, ContainerState } from "./constants";
 import { Executor } from "./executor";
 import { TelemetryClient } from "./telemetryClient";
 import { UserCancelledError } from "./UserCancelledError";
-import { create } from "domain";
 
 export class Utility {
     public static getConfiguration(): vscode.WorkspaceConfiguration {
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -15,6 +15,7 @@ import { Constants, ContainerState } from "./constants";
 import { Executor } from "./executor";
 import { TelemetryClient } from "./telemetryClient";
 import { UserCancelledError } from "./UserCancelledError";
+import { create } from "domain";
 
 export class Utility {
     public static getConfiguration(): vscode.WorkspaceConfiguration {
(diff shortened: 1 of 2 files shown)


src/common/utility.ts
```diff
@@ -475,17 +475,18 @@ export class Utility {
         return deployment;
     }
 
-    public static serializeCreateOptions(settings: any, createOptions: any): any {
+    public static serializeCreateOptions(moduleKey: string, settings: any, createOptions: any): any {
         let optionStr: string;
         if (typeof createOptions === "string") {
-            optionStr = createOptions;
+            // remove line breaks and extra spaces
+            optionStr = JSON.stringify(JSON.parse(createOptions), null, 0);
```
adashen: JSON parse may fail. Since now, the real createOptions will be json stringify and splitted into several chuncks (up to 8) in portal side. Like createOptions, createOptions01, createoptions02, createOptions03.....createOptions07 . So in this case if we just parse createOptions, it will fail.

src/common/utility.ts
```diff
@@ -475,17 +475,18 @@ export class Utility {
...
+            // remove line breaks and extra spaces
+            optionStr = JSON.stringify(JSON.parse(createOptions), null, 0);
         } else {
             optionStr = JSON.stringify(createOptions);
         }
         const re = new RegExp(`.{1,${Constants.TwinValueMaxSize}}`, "g");
         const options = optionStr.match(re);
         if (options.length > Constants.TwinValueMaxChunks) {
-            throw new Error("Size of createOptions too big. The maxium size of createOptions is 4K");
+            throw new Error(`Size of createOptions too big in module ${moduleKey}. The maximum size of createOptions is 4K`);
```
adashen: since we have settings in this method. We could get image name from settings. So no need to pass module.

adashen:
Thanks for the PR. And this let us know the issue when getting deployment manifest from portal

adashen:
@fbeltrao Can we give more detail step to copy the createOptions from portal? I just checked my edge device deployment on azure portal. And copy the createOptions from the last step. Seems there should not be so many white spaces.

fbeltrao:
@adashen the deployment.template.json file was created a few months ago, I had the impression it was created from the portal, I might be wrong.

You can see it here: 
https://github.com/Azure/iotedge-lorawan-starterkit/blob/master/LoRaEngine/deployment.template.json

adashen:
@fbeltrao Thanks for the information. I have tested your case and found it is an regex error in the code. I will open an issue and it will be fixed in the next release. Thank you very much for the information!

adashen:
issue https://github.com/Microsoft/vscode-azure-iot-edge/issues/278 opened

adashen:
issue #278 has been fixed in the latest release (v1.6.0) So close the PR here

PR #279, opened 2018-11-08
Remove ${} from completion items' sortText

LazarusX:
Order with this change:
image

### DIFF
--- a/src/intelliSense/configCompletionItemProvider.ts
+++ b/src/intelliSense/configCompletionItemProvider.ts
@@ -50,8 +50,9 @@ export class ConfigCompletionItemProvider implements vscode.CompletionItemProvid
         // }
 
         if (IntelliSenseUtility.locationMatch(location, Constants.imgDeploymentManifestJsonPath)) {
-            const images: string[] = await this.getSlnImgPlaceholders(document.uri);
-            return this.getCompletionItems(images, document, position, location);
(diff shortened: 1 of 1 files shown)


src/intelliSense/configCompletionItemProvider.ts
```diff
@@ -100,6 +100,14 @@ export class ConfigCompletionItemProvider implements vscode.CompletionItemProvid
             completionItem.range = overwriteRange;
             completionItem.insertText = value + separator;
             completionItem.kind = vscode.CompletionItemKind.Value;
+
+            // Deprioritize image placeholders with platform (e.g., ${MODULES.SampleModule.amd64}) in completion item list
```
adashen: what I am thinking is not deprioritize, but the correct group together of names. Can we use the value without "${}" for sort?

LazarusX: It's already grouped (master branch):
image

Do you mean we also need to remove  platforms like amd64 from sortText?

adashen: I mean sort the content in ${ }, MODULES.SampleFunction... Instead of the whole string  ${MODULES.SampleFunction....}  . use "split by ." is not robust. Since user may have folder name containing  "."

LazarusX:
New order:
image

PR #280, opened 2018-11-08
Default createOptions type is json object

adashen:
1. Make { } as default in deployment.template.json
2. object as default when adding module into deployment.template.json
3. Auto-complete use {} instead of  "{}" as default for createOptions

### DIFF
--- a/assets/solution/deployment.template.json
+++ b/assets/solution/deployment.template.json
@@ -17,7 +17,7 @@
             "type": "docker",
             "settings": {
               "image": "mcr.microsoft.com/azureiotedge-agent:1.0",
-              "createOptions": "{}"
+              "createOptions": {}
             }
           },
           "edgeHub": {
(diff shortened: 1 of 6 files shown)


LazarusX:
Please also update the module snippet: https://github.com/Microsoft/vscode-azure-iot-edge/blob/master/src/intelliSense/configCompletionItemProvider.ts#L22-L33

assets/solution/deployment.template.json
```diff
@@ -26,7 +26,15 @@
             "restartPolicy": "always",
             "settings": {
               "image": "mcr.microsoft.com/azureiotedge-hub:1.0",
-              "createOptions": "{\"HostConfig\":{\"PortBindings\":{\"5671/tcp\":[{\"HostPort\":\"5671\"}], \"8883/tcp\":[{\"HostPort\":\"8883\"}],\"443/tcp\":[{\"HostPort\":\"443\"}]}}}"
+              "createOptions": {
+                "HostConfig": {
+                  "PortBindings": {
+                    "5671/tcp": [{"HostPort": "5671"}],
```
LazarusX: Looks like the template here requires formatting.

It should be 
```
"HostConfig": {
                  "PortBindings": {
                    "8883/tcp": [
                      {
                        "HostPort": "8883"
                      }
                    ],
                    "443/tcp": [
                      {
                        "HostPort": "443"
                      }
                    ]
                  }
                }
```

LazarusX: Never mind. `JSON` will format it.

PR #284, opened 2018-11-14
Support build/generate/run customer deployment

adashen:
1. Build/Generate deployment now enabled for files with .template.json suffix. The context menu only support for our managed template file. But user could use command palette to choose their customized file.
2. Run now enabled for all .json files under config folder and the deployment.*.json
3. Change the managed debug template name to deployment.debug.template.json

### DIFF
--- a/src/container/containerManager.ts
+++ b/src/container/containerManager.ts
@@ -39,7 +39,7 @@ export class ContainerManager {
     }
 
     public async buildSolution(templateUri?: vscode.Uri, push: boolean = true, run: boolean = false): Promise<void> {
-        const pattern = `{${Constants.deploymentTemplatePattern},${Constants.debugDeploymentTemplatePattern}}`;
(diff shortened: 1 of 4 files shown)


src/container/containerManager.ts
```diff
@@ -143,6 +142,18 @@ export class ContainerManager {
         };
     }
 
+    private getDeployFileName(templateFileName: string, templateSchemaVersion: string): string {
+        const platform = templateSchemaVersion > "0.0.1" ? `.${Platform.getDefaultPlatform().platform}` : "";
```
formulahendry: String compare may have issue: e.g. "10.0" < "2.0"

adashen: yes. If we need version compare later, I will use utility to do so. Maybe from other existing version comparison npm. But for the current scenario, the compare works. Since the first char is "0".

PR #286, opened 2018-11-16
Add Azure Machine Learning integrations

LazarusX:


src/common/utility.ts
```diff
@@ -465,6 +475,19 @@ export class Utility {
         });
     }
 
+    public static getResourceGroupFromId(id: string): string {
```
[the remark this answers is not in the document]
LazarusX: Yes.

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
...
+        const result: HttpOperationResponse<IEndPoints> = await client.sendRequestWithHttpOperationResponse<IEndPoints>({
+            method: "GET",
+            url: workspace.discoveryUrl,
+            serializationMapper: undefined,
+            deserializationMapper: undefined,
+        });
+        if (result.response.statusCode === 200) {
+            return result.body.history;
+        } else {
+            throw new Error("API end points not found."); // TODO: return defaults by region
```
SLdragon: Do we need to show detailed error messages such as 5xx  errors

src/container/models/amlWorkspaceQuickPickItem.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as vscode from "vscode";
+import { AzureSubscription } from "../../typings/azure-account.api";
+
+export class AmlWorkspaceQuickPickItem implements vscode.QuickPickItem {
+    public readonly description: string;
+    public readonly detail?: string;
+
+    constructor(
+        public readonly label: string,
+        public readonly workspace: Workspace,
```
adashen: workspace passed in not consumed?

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+"use strict";
+import { AzureMachineLearningWorkspaces } from "azure-arm-machinelearningservices";
+import { Workspace } from "azure-arm-machinelearningservices/lib/models";
+import { Workspaces } from "azure-arm-machinelearningservices/lib/operations";
+import { HttpOperationResponse, ServiceClient, ServiceClientCredentials } from "ms-rest";
+import * as vscode from "vscode";
```
adashen: Seems import block is not ordered?

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        return imageItem.description;
+    }
+
+    private async selectWorkspace(): Promise<AmlWorkspaceQuickPickItem> {
+        await Utility.waitForAzLogin(this.azureAccount);
+
+        const workspaceItem: AmlWorkspaceQuickPickItem =
+            await vscode.window.showQuickPick(this.loadWorkspaceItems(), { placeHolder: "Select Azure Machine Learning Workspace", ignoreFocusOut: true });
```
adashen: Const?

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private async loadWorkspaceItems(): Promise<AmlWorkspaceQuickPickItem[]> {
+        try {
+            await this.azureAccount.waitForFilters();
+            const workspacePromises: Array<Promise<AmlWorkspaceQuickPickItem[]>> = [];
+            for (const azureSubscription of this.azureAccount.filters) {
+                const client: Workspaces = new AzureMachineLearningWorkspaces(
+                    azureSubscription.session.credentials,
+                    azureSubscription.subscription.subscriptionId!,
```
adashen: ! is for?

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
...
+            { placeHolder: "Select Image", ignoreFocusOut: true });
+        return imageItem;
+    }
+
+    private async loadAmlImageItems(workspace: Workspace, session: AzureSession): Promise<vscode.QuickPickItem[]> {
+        try {
+            const modelMgmtEndpoint: string = await this.getModelMgmtEndpoint(workspace);
+            const credentials: ServiceClientCredentials = session.credentials;
+            const client: ServiceClient = new ServiceClient(credentials);
+            const result: HttpOperationResponse<{ value: IImage[] }> = await client.sendRequestWithHttpOperationResponse<{ value: IImage[] }>({
```
adashen: So the lib didn't provide api to get images from workspace?

src/container/amlManager.ts
```diff
@@ -0,0 +1,154 @@
...
+            throw new Error("No available endpoint");
+        } else {
+            location = res[1];
+        }
+
+        return `https://${location}.modelmanagement.azureml.net`;
+    }
+
+    private async getHistoryEndpoint(workspace: Workspace): Promise<string> {
+        const client: ServiceClient = new ServiceClient();
```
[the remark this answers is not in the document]
LazarusX: Yes. The `discoveryUrl` is open to access. In my case, it's https://westus2.experiments.azureml.net/discovery.

src/common/utility.ts
adashen: also add undefined as return type

src/container/amlManager.ts
LazarusX: I don't quite get your point. Any more details?

src/common/utility.ts
LazarusX: Will add.

src/container/amlManager.ts
LazarusX: Just a hint to tell the compiler that the `subscriptionId` cannot be null, which can be useful with some strict compiler flag. See https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator. This is copied from the sample code https://github.com/Microsoft/vscode-azure-account/blob/master/sample/src/extension.ts#L78

src/container/acrManager.ts
```diff
@@ -86,37 +80,27 @@ export class AcrManager {
...
-                const client = new ContainerRegistryManagementClient(
+                const client: Registries = new ContainerRegistryManagementClient(
                     azureSubscription.session.credentials,
                     azureSubscription.subscription.subscriptionId!,
-                );
+                ).registries;
 
                 registryPromises.push(
-                    this.listAcrRegistries(client.registries, client.registries.list())
+                    Utility.listAzureResources<Registry>(client.list(), client.listNext)
```
adashen: Does azure resrouces (Registries, Workspaces) share the same interface? If so, the input of listAzureResources only need the interface?

src/container/amlManager.ts
```diff
@@ -0,0 +1,159 @@
...
+                return { label: image.name, description: image.imageLocation };
+            });
+        } catch (error) {
+            error.message = `Error fetching image list: ${error.message}`;
+            throw error;
+        }
+    }
+
+    private getWorkspaceNameCounts(workspaces: Workspace[]): Map<string, number> {
+        const counts: Map<string, number> = new Map<string, number>();
```
adashen: the name of workspaces could be duplicated under ml account?

LazarusX: Yes. I have verified and comment is added in https://github.com/Microsoft/vscode-azure-iot-edge/pull/286/files#diff-7dff5585ca5daca8278354dee63e24b3R58.

src/container/amlManager.ts
LazarusX: This code snippet is copied from AI tools. Given we have little documentation about the API, I tend to keep it as is.

src/container/amlManager.ts
```diff
@@ -0,0 +1,159 @@
...
+        const historyEndpoint: string = await this.getHistoryEndpoint(workspace);
+        let location: string;
+        const res: string[] = historyEndpoint.match(/\/\/(.*?)\./);
+        if (res.length < 2) {
+            throw new Error("No available endpoint");
+        } else {
+            location = res[1];
+        }
+
+        return `https://${location}.modelmanagement.azureml.net`;
```
adashen: why not just get the modelmanagement from response instead of calculate base don location? It may cause potential error if it is not global azure

LazarusX: Code in this method is copied from AI tools, and I cannot find any documentation about  the REST API of Azure Machine Learning Model Management services. There might be some edge cases where we cannot simply get the model management endpoint from the discovery URL.

PR #302, opened 2018-11-21
Refine default platform place holder

adashen:


src/edge/edgeManager.ts
```diff
@@ -193,7 +193,7 @@ export class EdgeManager {
             platformMap.set(displayName, value);
         });
         const platformNames: string[] = (keyWithAlias.sort()).concat(defaultKeys.sort());
-        const defaultPlatform = await vscode.window.showQuickPick(platformNames, { placeHolder: Constants.selectPlatform, ignoreFocusOut: true });
+        const defaultPlatform = await vscode.window.showQuickPick(platformNames, { placeHolder: Constants.selectDefaultPlatform, ignoreFocusOut: true });
```
LazarusX: Look like there is another `selectPlatform` occurrence.

adashen: Another one is used when build module. It should not be changed

PR #306, opened 2018-11-29
support setting versions

adashen:


package.json
```diff
@@ -186,6 +186,61 @@
...
+        },
+        "azure-iot-edge.version.tempSensor": {
+          "type": "string",
+          "default": "1.0",
+          "description": "Set the tempSensor image version, which will be referenced in deployment manifest."
+        },
+        "azure-iot-edge.version.cmodule": {
+          "type": "string",
+          "default": null,
+          "description": "Set the template version of C module template, which will be refernced during scaffolding C module. null for latest."
```
LazarusX: `Set the template version of C module template` -> `Set the C module template version` for simplicity and alignment with image version. Same for other templates.

package.json
```diff
@@ -186,6 +186,61 @@
...
+          "type": "string",
+          "default": null,
+          "description": "Set the C# fuction module template version, which will be refernced during scaffolding C# function module. null for latest."
+        },
+        "azure-iot-edge.version.javamodule": {
+          "type": "string",
+          "default": null,
+          "description": "Set the Java module template version, which will be refernced during scaffolding Java module. null for latest."
+        },
+        "azure-iot-edge.version.nodemodule": {
```
LazarusX: Node.js module template version does not work because users manage them outside VS Code.

src/edge/edgeManager.ts
```diff
@@ -404,16 +408,24 @@ export class EdgeManager {
...
+                    const csfuncversion = Versions.csFunctionTemplateVersion();
+                    const installCmd = csfuncversion != null ? `new -i Microsoft.Azure.IoT.Edge.Function::${csfuncversion}` : "new -i Microsoft.Azure.IoT.Edge.Function";
+                    await Executor.executeCMD(outputChannel, "dotnet", { shell: true }, installCmd);
+                }
                 await Executor.executeCMD(outputChannel, "dotnet", { cwd: `${parent}`, shell: true }, `new aziotedgefunction -n "${name}" -r ${repositoryName}`);
                 break;
             case Constants.LANGUAGE_PYTHON:
                 const gitHubSource = "https://github.com/Azure/cookiecutter-azure-iot-edge-module";
-                const branch = "master";
+                const branch = Versions.pythonTemplateVersion();
```
LazarusX: The default branch of https://github.com/Azure/cookiecutter-azure-iot-edge-module is develop, so if branch is undefined, users will get develop template.

src/edge/edgeManager.ts
```diff
@@ -424,7 +436,7 @@ export class EdgeManager {
                 break;
             case Constants.LANGUAGE_C:
                 await new Promise((resolve, reject) => {
-                    download("github:Azure/azure-iot-edge-c-module#master", path.join(parent, name), (err) => {
+                    download(`github:Azure/azure-iot-edge-c-module#${Versions.cTemplateVersion()}`, path.join(parent, name), (err) => {
```
LazarusX: Similar issue with Python template. We should explicitly fall back to master when cTemplateVersion is null.

PR #323, opened 2018-12-12
Use Webpack to improve extension performance

LazarusX:


.vscode/launch.json
```diff
@@ -7,22 +7,29 @@
...
-            "sourceMaps": true,
-            "outFiles": [ "${workspaceRoot}/out/src/**/*.js" ],
-            "preLaunchTask": "npm"
+            "args": [
+                "--extensionDevelopmentPath=${workspaceFolder}"
+            ],
+            "outFiles": [
+                "${workspaceFolder}/dist/**/*.js"
+            ],
+            "preLaunchTask": "npm: webpack"
```
adashen: webpack or webpack-dev?

.vscode/launch.json
```diff
@@ -7,22 +7,29 @@
...
+                "--extensionTestsPath=${workspaceRoot}/out/test"
+            ],
             "stopOnEntry": false,
             "sourceMaps": true,
-            "outFiles": [ "${workspaceRoot}/out/test/**/*.js" ],
-            "preLaunchTask": "npm"
+            "outFiles": [
+                "${workspaceRoot}/out/test/**/*.js"
+            ],
+            "preLaunchTask": "npm: compile"
```
adashen: why remove the npm task? It has some problem watch check.

package.json
```diff
@@ -411,11 +411,13 @@
     ]
   },
   "scripts": {
-    "vscode:prepublish": "tsc -p ./",
-    "compile": "tsc -watch -p ./",
+    "vscode:prepublish": "webpack --mode production",
+    "compile": "tsc -p ./",
```
adashen: why remote watch mode?

.vscode/launch.json
LazarusX: VS Code reports schema deprecation warning of current task.json. After removing task.json, VS Code can automatically discovery npm scripts as tasks and we don't need to configure them manually anymore.

.vscode/launch.json
LazarusX: `webpack-dev` will never return so the extension host cannot start.

package.json
adashen: it is used in "launch tests"?

.vscode/launch.json
adashen: so what is webpack-dev used for?

LazarusX: It's copied from the sample, and not used currently.

package.json
LazarusX: Yes, that's also the use case of CI.

adashen: So now the test is not compiled in watch mode. Any impact? Will the reload still work?

LazarusX: No. We seldom run the test launch profile locally in watch mode in VS Code, as VS Code does not allow running the test with any VS Code instance open.

PR #331, opened 2018-12-24
add telemetry to track simulator auto install

adashen:


src/edge/simulator.ts
```diff
@@ -56,10 +63,14 @@ export class Simulator {
...
 
     private static async validateSimulatorInstalled(outputChannel: vscode.OutputChannel = null): Promise<InstallReturn> {
+        const telemetryName = "simulatorInstalled";
         if (await Simulator.simulatorInstalled()) {
             return InstallReturn.Success;
         } else {
-            return await Simulator.installSimulatorWithPip(true, Constants.needSimulatorInstalledMsg, outputChannel);
+            TelemetryClient.sendEvent(`${telemetryName}.install`);
+            const installRes = await Simulator.installSimulatorWithPip(true, Constants.needSimulatorInstalledMsg, outputChannel);
+            TelemetryClient.sendEvent(`${telemetryName}.${InstallReturn[installRes]}`);
```
formulahendry: ${telemetryName}.install.${InstallReturn[installRes]} ?

adashen: telemetryName is already contains install. It only sent when install needed.

src/edge/simulator.ts
```diff
@@ -38,16 +41,20 @@ export class Simulator {
             } else {
                 message = Constants.updateSimulatorMsg;
             }
+            type = "upgrade";
         } catch (error) {
             message =  Constants.needSimulatorInstalledMsg;
+            type = "install";
         }
 
         if (message) {
```
xinyiz1226: Need to check type value here
```suggestion
        if (message) {
```

adashen: type is the telemetry name. message has value means we need to install or update the iotedgehubdev tool. I will add default value to type as empty string to be safer.

src/edge/simulator.ts
adashen: will update to be consistent with update scenario, Thx

formulahendry: So ``${telemetryName}.install`` could be ``${telemetryName}``?

adashen: updated.

src/edge/simulator.ts
adashen: code updated

PR #333, opened 2018-12-29
Use git download instead of cookiecutter to add Python module

formulahendry:


package.json
```diff
@@ -460,8 +461,9 @@
     "jsonc-parser": "^1.0.1",
     "request": "^2.88.0",
     "request-promise": "^4.2.2",
-    "strip-json-comments": "^2.0.1",
     "semver": "^5.6.0",
+    "strip-json-comments": "^2.0.1",
+    "tmp": "0.0.33",
```
adashen: use ^ to match the latest minor release?

package.json
```diff
@@ -435,9 +435,10 @@
     "@types/mocha": "^2.2.32",
     "@types/node": "^6.0.40",
     "@types/request-promise": "^4.1.41",
+    "@types/semver": "^5.5.0",
     "@types/sinon": "^5.0.5",
     "@types/strip-json-comments": "0.0.30",
-    "@types/semver": "^5.5.0",
+    "@types/tmp": "0.0.33",
```
adashen: use ^ to match the latest minor release?

src/edge/edgeManager.ts
```diff
@@ -397,12 +398,25 @@ export class EdgeManager {
...
                 break;
             case Constants.LANGUAGE_PYTHON:
-                const gitHubSource = "https://github.com/Azure/cookiecutter-azure-iot-edge-module";
-                const branch = Versions.pythonTemplateVersion();
-                await Executor.executeCMD(outputChannel,
-                    "cookiecutter",
-                    { cwd: `${parent}`, shell: true },
-                    `--no-input ${gitHubSource} module_name=${name} image_repository=${repositoryName} --checkout ${branch}`);
+                await new Promise((resolve, reject) => {
+                    tmp.dir({ unsafeCleanup: true }, (err, tmpDir, cleanupCallback) => {
```
adashen: Why not just download to the module path? 
The benefit to use temp folder, is if error happens, no pollution to the project folder. If that is the intent, how about replace the repositoryname in the temp folder?

package.json
formulahendry: ^0.0.33 := >=0.0.33 <0.0.34   
So no need to use ^

package.json
formulahendry: ^0.0.33 := >=0.0.33 <0.0.34   
So no need to use ^

adashen: so we don't want to keep updated with minor version?  for example later it has 0.1.x?

src/edge/edgeManager.ts
formulahendry: There is no API to download a folder of a repo, so we use tmp folder.

package.json
formulahendry: `0.0.5 -> 0.1.0` would be a little  different with `1.0.5 -> 1.1.0`. `0.0.5 -> 0.1.0` may contains breaking change. And `0.0.33` is auto generated by `npm install`, I think that is the reason

src/edge/edgeManager.ts
```diff
@@ -397,12 +398,30 @@ export class EdgeManager {
...
+                            download(`github:Azure/cookiecutter-azure-iot-edge-module#${Versions.pythonTemplateVersion()}`, tmpDir, async (downloadErr) => {
+                                if (downloadErr) {
+                                    reject(downloadErr);
+                                } else {
+                                    try {
+                                        await this.updateRepositoryName(tmpDir, moduleContentDirName, repositoryName);
+                                        await fse.move(moduleContentDir, path.join(parent, name));
+                                        resolve();
+                                    } catch (error) {
+                                        reject(err);
```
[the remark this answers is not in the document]
formulahendry: Code will still be executed after `resolve()` or `reject()`. The the final `cleanupCallback` will be always executed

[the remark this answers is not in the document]
adashen: better to add finally block for clean code.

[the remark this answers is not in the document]
formulahendry: `Finally` is useful when there is possibilty that error is thrown from the catch block. Not match with our case, And the `cleanupCallback();` is used in a bigger scope -- the `download` function code block

Issue #324, opened 2018-12-14
Error: cannot read property 'document' of undefined when swith default platform


adashen:
The error will show when there is no file open in the workspace

adashen:
@czgtest please verify the fix with 1.9.0-rc

adashen:
@czgtest please verify the fix with 1.9.0-rc

PR #349, opened 2019-01-18
suppor env for created terminal and subprocess and fix issue #324 and…

adashen:
… turn NSAT survey to 100%

### DIFF
--- a/src/common/executor.ts
+++ b/src/common/executor.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT license.
 
 "use strict";
-import { ChildProcess, exec, execSync, ExecSyncOptions, spawn, SpawnOptions } from "child_process";
+import { ChildProcess, execSync, ExecSyncOptions, spawn, SpawnOptions } from "child_process";
 import * as vscode from "vscode";
 import { Configuration } from "./configuration";
 import { Constants } from "./constants";
--- a/src/common/executor.ts
+++ b/src/common/executor.ts
@@ -2,25 +2,27 @@
 // Licensed under the MIT license.
 
 "use strict";
-import { ChildProcess, exec, execSync, spawn, SpawnOptions } from "child_process";
+import { ChildProcess, exec, execSync, ExecSyncOptions, spawn, SpawnOptions } from "child_process";
 import * as vscode from "vscode";
+import { Configuration } from "./configuration";
 import { Constants } from "./constants";
 
 export class Executor {
     public static runInTerminal(command: string, terminal: string = Constants.edgeDisplayName): void {
         if (this.terminals[terminal] === undefined) {
-            this.terminals[terminal] = vscode.window.createTerminal(terminal);
+            this.terminals[terminal] = Executor.createTerminal(terminal);
         }
         this.terminals[terminal].show();
         this.terminals[terminal].sendText(command);
     }
 
-    public static exec(command: string) {
-        return exec(command);
-    }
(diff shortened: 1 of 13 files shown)


src/common/executor.ts
```diff
@@ -2,25 +2,27 @@
 // Licensed under the MIT license.
 
 "use strict";
-import { ChildProcess, exec, execSync, spawn, SpawnOptions } from "child_process";
+import { ChildProcess, exec, execSync, ExecSyncOptions, spawn, SpawnOptions } from "child_process";
```
LazarusX: Minor: `exec` is imported but not used now.

adashen: Thanks, will remove

src/common/executor.ts
```diff
@@ -78,4 +87,34 @@ export class Executor {
             outputPane.appendLine(value);
         }
     }
+
+    private static createTerminal(terminal: string): vscode.Terminal {
+        const envVars  = Executor.getEnvFromConfig();
+        const options: vscode.TerminalOptions = { name: terminal };
```
formulahendry: Would `const options: vscode.TerminalOptions = { name: terminal, env:  envVars};` be simpler?

src/common/executor.ts
```diff
@@ -2,25 +2,27 @@
...
     }
 
-    public static exec(command: string) {
-        return exec(command);
-    }
-
     public static execSync(command: string) {
-        return execSync(command, { encoding: "utf8" });
+        const envVars = Executor.getEnvFromConfig();
+        const options: ExecSyncOptions = { encoding: "utf8" };
```
formulahendry: Same here: `{ encoding: "utf8", env:  Executor.mergeEnvs(envVars, process.env)}`

adashen: This will add an key with undefined value. It is different from there is no key.

src/common/executor.ts
adashen: Same as above.

formulahendry: Per definition, env parameter is `Object with environment variables that will be added to the VS Code process.` It is added values, it would not affect other environment variables

adashen: Thanks, got it. To make it consistent in the code as subprocess scenario. I prefer to add only valid key which is more safer.

src/common/executor.ts
```diff
@@ -35,6 +37,13 @@ export class Executor {
 
             let stderr: string = "";
             let stdOutput: string = "";
+            const envVars = Executor.getEnvFromConfig();
+            if (envVars) {
+                options = options || {};
+                const processEnvs = Executor.mergeEnvs(envVars, process.env);
+                options.env = Executor.mergeEnvs(options.env, processEnvs);
```
LazarusX: Just confirm, the priority order of envs is: `options.env` > envs from config > `process.env`?

adashen: Yes. options.env is passed by caller with intend. Envs from config is an adds-on to the vs code process.

src/common/executor.ts
```diff
@@ -78,4 +87,34 @@ export class Executor {
             outputPane.appendLine(value);
         }
     }
+
+    private static createTerminal(terminal: string): vscode.Terminal {
+        const envVars  = Executor.getEnvFromConfig();
```
LazarusX: Do we need to merge `envVars` with `process.env`?

PR #353, opened 2019-01-28
avoid override process env

adashen:


src/common/executor.ts
```diff
@@ -20,7 +20,8 @@ export class Executor {
         const envVars = Executor.getEnvFromConfig();
         const options: ExecSyncOptions = { encoding: "utf8" };
         if (envVars) {
-            options.env = Executor.mergeEnvs(envVars, process.env);
+            const processEnvs = JSON.parse(JSON.stringify(process.env));
+            options.env = Executor.mergeEnvs(envVars, processEnvs);
```
LazarusX: How about putting the clone process in `mergeEnvs` and return the merged env?

LazarusX: Currently, `mergeEnvs` merge envs in place and the second parameter is also altered.

PR #365, opened 2019-02-13
fix #364

### LINKED ISSUE #364: Set Default Platform: Error: Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again.
- VSCode Version: 1.30.1
- OS Version:
- Extension Version: 1.9.0

Steps to Reproduce:

1. Close workspace in vscode.
2. Use command palette to call "Set Default Target Platform for Edge Solution"
3. Error message will show "Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again."

adashen:


LazarusX:
Can we reuse https://github.com/Microsoft/vscode-azure-iot-edge/blob/master/src/common/utility.ts#L22-L29?

src/common/utility.ts
```diff
@@ -19,9 +19,10 @@ import { TelemetryClient } from "./telemetryClient";
 import { UserCancelledError } from "./UserCancelledError";
 
 export class Utility {
-    public static checkWorkspace(): boolean {
+    public static checkWorkspace(message?: string): boolean {
```
LazarusX: Nit: `message: string = Constants.noWorkspaceMsg`

Issue #379, opened 2019-03-05
Open gallery sample project name verification is not correct


SLdragon:
"vision-ai-developer-kit" is a valid name, however, the extension shows that we cannot use this name.

![image](https://user-images.githubusercontent.com/5545529/53789887-7b7e7f80-3f1d-11e9-8dd3-6db750d95c89.png)

SLdragon:
Fixed and close

Issue #76, opened 2018-03-20
[Add IoT Edge Module] if the module name contains double quotes,will get error.


czgtest:
if the module name is "c#module03", and  localhost:5000/"c#module03"

![image](https://user-images.githubusercontent.com/35018765/37644636-675ae092-2c5f-11e8-820c-4d718ee99bb4.png)

![image](https://user-images.githubusercontent.com/35018765/37644749-bc573f78-2c5f-11e8-8b12-aecac7d1fb5f.png)

![image](https://user-images.githubusercontent.com/35018765/37644574-3455beb0-2c5f-11e8-9bb5-a310e3ff7fb5.png)

Verify the deployment.template.json file is not updated: the module JSON is not added into the modules section

adashen:
@LazarusX  please have a look. Actually this is a user error to input invalid name. But we may find a better way to report the issue to user.

formulahendry:
For #74, #75, #76,  I suggest check the name when user type in the input box and only allow alphanumeric characters or the symbol _

adashen:
@czgtest Please help to verify the fix

czgtest:
Now solution/module name only allow alphanumeric characters .
![image](https://user-images.githubusercontent.com/35018765/37747842-1155aef8-2dbc-11e8-8d4a-b6adf5e0440e.png)

Issue #75, opened 2018-03-20
[New IoT Edge solution]Do we allow Solution /Module Name to be  spaces ?


czgtest:
![image](https://user-images.githubusercontent.com/35018765/37643833-c6472834-2c5c-11e8-80ca-c360e734c2d0.png)

adashen:
@LazarusX it seems similar to another naming issue. So also assigned to you.

adashen:
@czgtest please verify

czgtest:
now if enter space key to solution/Module name,  it doesn't allow now.

![image](https://user-images.githubusercontent.com/35018765/37747627-212296da-2dbb-11e8-9b7b-5c453c8c14c1.png)

Issue #74, opened 2018-03-20
[New IoT Edge solution] if input module name is  /Sample/Module/, will get error


czgtest:
if input module name is  /Sample/Module/, will get error : 
Command failed with exit code 214750003

czgtest:
another issue 

if input module name is  Edge:Solution,  will get error too.

![image](https://user-images.githubusercontent.com/35018765/37644049-74c72e0e-2c5d-11e8-8ad2-fb2b1f7f2c5f.png)

adashen:
@czgtest please verify the fix

czgtest:
fix work now,  if input module name is /Sample/Module/  or Edge:Solution,  it will reminder user to input alphanumeric characters

![image](https://user-images.githubusercontent.com/35018765/37747416-1489037e-2dba-11e8-9d9a-2ddca17a85c8.png)


![image](https://user-images.githubusercontent.com/35018765/37747520-9769d228-2dba-11e8-9905-2414faee9f8d.png)

PR #380, opened 2019-03-05
Fix issue when check sample solution name

SLdragon:
https://github.com/Microsoft/vscode-azure-iot-edge/issues/379

### DIFF
--- a/src/edge/edgeManager.ts
+++ b/src/edge/edgeManager.ts
@@ -609,6 +609,22 @@ export class EdgeManager {
         return undefined;
     }
 
+    private async validateSolutionName(name: string, parentPath?: string): Promise<string | undefined> {
+        if (!name) {
+            return "The name could not be empty";
+        }
+        if (name.match(/[/\\:*?\"<>|]/)) {
+            return "Solution name cannot contain special characters";
+        }
+        if (parentPath) {
+            const folderPath = path.join(parentPath, name);
+            if (await fse.pathExists(folderPath)) {
+                return `${name} already exists under ${parentPath}`;
+            }
+        }
+        return undefined;
+    }
+
     private validateModuleExistence(name: string, modules?: string[]): string | undefined {
         if (modules && modules.indexOf(name) >= 0) {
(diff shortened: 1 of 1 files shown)


src/edge/edgeManager.ts
```diff
@@ -609,6 +609,22 @@ export class EdgeManager {
         return undefined;
     }
 
+    private async validateSolutionName(name: string, parentPath?: string): Promise<string | undefined> {
+        if (!name) {
+            return "The name could not be empty";
+        }
+        if (name.match(/[/\\:*?\"<>|]/)) {
+            return "Solution name cannot contain special characters";
```
LazarusX: "Special characters" here is not clear enough. I prefer something like "Only *** are allowed".

SLdragon: User can use Chinese or other utf8 characters for folder names, so I think "Only..." is not easy to describe the limitation

src/edge/edgeManager.ts
```diff
@@ -609,6 +609,22 @@ export class EdgeManager {
         return undefined;
     }
 
+    private async validateSolutionName(name: string, parentPath?: string): Promise<string | undefined> {
+        if (!name) {
```
adashen: name could be spaces? So we need to trim the spaces

src/edge/edgeManager.ts
adashen: How about change the message to Solution name can't contain characters: \/:*?<>|

src/edge/edgeManager.ts
SLdragon: Oh, yes, I will trim it.

src/edge/edgeManager.ts
SLdragon: Thanks, fixed

SLdragon:
Also verified
https://github.com/Microsoft/vscode-azure-iot-edge/issues/74
https://github.com/Microsoft/vscode-azure-iot-edge/issues/75
https://github.com/Microsoft/vscode-azure-iot-edge/issues/76

PR #384, opened 2019-03-07
Add module name check logic

SLdragon:


src/marketplace/localserver.ts
```diff
@@ -59,6 +63,16 @@ export class LocalServer {
         });
     }
 
+    private async isValidModuleName(req: express.Request, res: express.Response, next: express.NextFunction) {
```
formulahendry: better naming: validateModuleName

PR #393, opened 2019-03-12
Populate PlanId Env when adding ASA module

SLdragon:


src/edge/edgeManager.ts
```diff
@@ -728,7 +730,7 @@ export class EdgeManager {
             debugImageName = debugSettings.debugImageName;
             debugCreateOptions = debugSettings.debugCreateOptions;
         }
-        return new ModuleInfo(module, repositoryName, imageName, moduleTwin, createOptions, debugImageName, debugCreateOptions);
+        return new ModuleInfo(module, repositoryName, imageName, moduleTwin, createOptions, debugImageName, debugCreateOptions, [], env);
```
adashen: if null, will then env block generated in json?

PR #395, opened 2019-03-12
Marketplace enhancement

formulahendry:


src/marketplace/marketplace.ts
```diff
@@ -64,6 +64,7 @@ export class Marketplace {
             const moduleInfo = new ModuleInfo(message.moduleName, repositoryName, message.imageName, message.twins, message.createOptions,
                 message.imageName, message.createOptions, message.routes, message.environmentVariables);
             await vscode.commands.executeCommand("azure-iot-edge.internal.addModule", this.templateFile, this.isNewSolution, moduleInfo, Constants.MARKETPLACE_MODULE);
+            TelemetryClient.sendEvent("addMarketplaceModule", { moduleId: message.id });
```
adashen: move before command execution? So that even if the command fails. telemetry still be sent?

formulahendry: Moved

Issue #410, opened 2019-03-18
error handling when ACR don't have image


czgtest:
- VSCode Version:1.32.3
- OS Version: Windows10
- Extension Version: 1.11.0-rc2

Steps to Reproduce:

1.create new solution for existing module(import for ACR)
2.select one Azure Container Registry which don't have any image
get Error fetching repository list: Cannot read property 'map' of null
Expect:
give some error handling and reminder user there is no image of ACR
![image](https://user-images.githubusercontent.com/35018765/54520611-1195b980-49a4-11e9-8993-33f9b6fb85ec.png)

czgtest:
fix work， if I select an ACR without any repository, it will pop up an error message to give user reminder
![image](https://user-images.githubusercontent.com/35018765/54732221-9283c980-4bcd-11e9-9458-8d266cde3dfb.png)

PR #412, opened 2019-03-19
show warning message when ACR don't have image

SLdragon:
Fix issue : https://github.com/Microsoft/vscode-azure-iot-edge/issues/410

### DIFF
--- a/src/container/acrManager.ts
+++ b/src/container/acrManager.ts
@@ -103,16 +103,12 @@ export class AcrManager {
     }
 
     private async selectAcrRepo(registryUrl: string, session: AzureSession): Promise<vscode.QuickPickItem> {
-        const repoItems = await this.loadAcrRepoItems(registryUrl, session);
-        if (!repoItems) {
-            vscode.window.showWarningMessage("There is no repository in the registry.");
-            return;
-        }
-        const acrRepoItem: vscode.QuickPickItem = await vscode.window.showQuickPick(repoItems, { placeHolder: "Select Repository", ignoreFocusOut: true });
+        const acrRepoItem: vscode.QuickPickItem = await vscode.window.showQuickPick(this.loadAcrRepoItems(registryUrl, session), { placeHolder: "Select Repository", ignoreFocusOut: true });
         return acrRepoItem;
     }
 
     private async loadAcrRepoItems(registryUrl: string, session: AzureSession): Promise<vscode.QuickPickItem[]> {
+        let isEmptyRegistry = false;
         try {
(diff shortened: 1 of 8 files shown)


src/container/acrManager.ts
```diff
@@ -121,6 +125,12 @@ export class AcrManager {
 
             const repoItems: vscode.QuickPickItem[] = [];
             const repos = JSON.parse(catalogResponse).repositories;
+
+            if (!repos) {
+                vscode.window.showWarningMessage("There is no repository in the registry.");
+                return;
```
adashen: return null is better?

src/container/acrManager.ts
```diff
@@ -103,7 +103,12 @@ export class AcrManager {
     }
 
     private async selectAcrRepo(registryUrl: string, session: AzureSession): Promise<vscode.QuickPickItem> {
-        const acrRepoItem: vscode.QuickPickItem = await vscode.window.showQuickPick(this.loadAcrRepoItems(registryUrl, session), { placeHolder: "Select Repository", ignoreFocusOut: true });
+        const repoItems = await this.loadAcrRepoItems(registryUrl, session);
+        if (!repoItems) {
+            vscode.window.showWarningMessage("There is no repository in the registry.");
+            return;
+        }
+        const acrRepoItem: vscode.QuickPickItem = await vscode.window.showQuickPick(repoItems, { placeHolder: "Select Repository", ignoreFocusOut: true });
```
LazarusX: FYI, by updating the parameter passed to `showQuickPick` from a `Promise` to an `array`, VS Code will not show loading animations.

adashen: Good catch. @SLdragon, how about keep "promise", and return empty list of quickpickitem when there is no repository

SLdragon:
Updated
empty-repository

src/container/acrManager.ts
LazarusX: Another option is to simply throw a 404 error and https://github.com/Microsoft/vscode-azure-iot-edge/pull/412/files#diff-36614ddf00470008048ba066990c9c6fR145 will handle it.

src/container/acrManager.ts
```diff
@@ -121,6 +122,12 @@ export class AcrManager {
 
             const repoItems: vscode.QuickPickItem[] = [];
             const repos = JSON.parse(catalogResponse).repositories;
+
+            if (!repos) {
+                isEmptyRegistry = true;
+                throw new Error("There is no repository in the registry.");
```
formulahendry: `"There is no repository in the registry."`  is this used?

SLdragon: yes, it is used in the error message

PR #413, opened 2019-03-19
do not allow module name start with number

SLdragon:


src/common/utility.ts
```diff
@@ -566,6 +566,10 @@ export class Utility {
         if (name.match(/[^a-zA-Z0-9\_]/)) {
             return "The name must contain only alphanumeric characters or the symbol _";
         }
+        if (name.match(/^\d/)) {
```
adashen: instead of do regex match two times. Can we just match with the legal pattern once?

SLdragon: Yes, however, if we use one pattern, the error message maybe not accurate

adashen: How about error message: "The name cannot start with number and must contain  only alphanumeric characters or the symbol _"

SLdragon: Thanks, according to the schema, the validate regex would be /^[a-zA-Z][a-zA-Z0-9_.-]*$/

SLdragon:
We have already updated the json schema store to support  module name start with number, so close this pull request

PR #424, opened 2019-03-30
fix marketplace icon parse issue

SLdragon:


src/marketplace/localserver.ts
```diff
@@ -89,20 +89,22 @@ export class LocalServer {
...
-                        item.iotEdgeMetadataUrl = metaData.uri;
-                    } else {
-                        item.iotEdgeMetadataUrl = null;
-                    }
+                    if (item.plans && item.plans[0] && item.plans[0].artifacts) {
+                        const metaData = item.plans[0].artifacts.find((artifact) => artifact.name === "iot-edge-metadata.json");
+                        if (metaData !== null) {
+                            item.iotEdgeMetadataUrl = metaData.uri;
+                        } else {
+                            item.iotEdgeMetadataUrl = null;
```
adashen: what will happen if matadata url is null? although not related to the hot fix. This maybe an issue we need take care.

src/marketplace/localserver.ts
```diff
@@ -89,20 +89,22 @@ export class LocalServer {
...
+                        }
 
-                    const image = item.images[0].items.find((img) => img.id === "small");
-                    if (image !== null) {
-                        item.icon = image.uri;
-                    } else {
-                        item.icon = null;
+                        const iconFileUris = item.iconFileUris;
```
formulahendry: This code block should not be inside `if (item.plans && item.plans[0] && item.plans[0].artifacts)` block

SLdragon: If we cannot parse plans, do we need to skip this module?

adashen: If we could not get metadata, the current code will get error when user click import right? If so, I think we may not put the invalid modules in to item list

SLdragon: Ok, if we cannot find the meta url, then server side will not send this module to front end

Issue #178, opened 2018-06-23
Docker/moby requires admin privileges on ubuntu


shizn:
it looks like the extension expects the current user to be able to run docker commands. turns out the default docker/moby installed requires admin privileges on ubuntu. either the extension should be able to sudo itself around (prompting for a password) or the error should be a little more helpful and our docs should reference a tutorial on how to enable that.
something like this: https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user

jantielens:
I just bumped into this issue as well (while trying out the new VS Code Remote Development features). So I guess it will be much more common in the future to have this problem. The link you provided is indeed the correct solution IMHO. I've created a pull request to add both a FAQ and prerequisite related to this problem. #438

PR #438, opened 2019-05-07
FAQ & prereq for assumed managing docker perms

jantielens:
The extension assumes that the current user has the permissions to manage docker. On Linux this is not the case by default. I've added a FAQ item and a prerequisite item to explain this, and the commands to assign these permissions on Linux (including a link to the official Docker docs).

Background: on Windows (where you typically run VS code), by default the current user will have sufficient permissions. But with the new VS Code Remote Development tooling (https://code.visualstudio.com/docs/remote/remote-overview) it will be much more common that the extension is actually running on a Linux machine (while VS Code itself is still running on Windows). This is exactly how I bumped into this issue. :)


jantielens:
There is an open issue related to this problem: #178

shizn:
Thanks for your contribution!!!

PR #440, opened 2019-05-13
Support MODULEDIR schema

adashen:
1. Upgrade the version of new create deployment template to 2.0.0
2. Parse MODULEDIR<> schema to get the external module.json
3. Clean the logic to support submodule/external module

### DIFF
--- a/assets/solution/deployment.template.json
+++ b/assets/solution/deployment.template.json
@@ -1,5 +1,5 @@
 {
-  "$schema-template": "1.0.0",
+  "$schema-template": "2.0.0",
   "modulesContent": {
     "$edgeAgent": {
       "properties.desired": {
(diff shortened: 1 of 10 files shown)


assets/solution/deployment.template.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema-template": "1.0.0",
+  "$schema-template": "2.0.0",
```
SLdragon: Is this version related to the schemastore json schema version? Now our latest version is 1.1.0

formulahendry:
Shall we also add IntelliSense for MODULEDIR?
image

adashen:
> Shall we also add IntelliSense for MODULEDIR?
> image

Since we have no control of external module path. So it is not shown. However, if the user has already added external MODULEDIR, it will be shown here.

assets/solution/deployment.template.json
adashen: Discussed offline. Schemastore will have a new version.

PR #441, opened 2019-05-14
Update json schema version to 2.0

SLdragon:


package.json
```diff
@@ -297,11 +297,11 @@
     "jsonValidation": [
       {
         "fileMatch": "deployment.template.json",
-        "url": "http://json.schemastore.org/azure-iot-edge-deployment-template-1.0"
+        "url": "http://json.schemastore.org/azure-iot-edge-deployment-template-2.0"
```
adashen: So that means VSCode has capability to handle both draft04/draft07?

SLdragon: Yes, it support both 04 and 07

PR #458, opened 2019-06-12
Show error message when scaffolding fail

adashen:


src/edge/edgeManager.ts
```diff
@@ -448,96 +449,120 @@ export class EdgeManager {
...
+                try {
+                    if (Versions.installCSharpTemplate()) {
+                        const csversion = Versions.csTemplateVersion();
+                        const installCmd = csversion != null ? `new -i Microsoft.Azure.IoT.Edge.Module::${csversion}` : "new -i Microsoft.Azure.IoT.Edge.Module";
+                        await Executor.executeCMD(outputChannel, "dotnet", { shell: true }, installCmd);
+                    }
+                    await Executor.executeCMD(outputChannel, "dotnet", { cwd: `${parent}`, shell: true }, `new aziotedgemodule -n "${name}" -r ${repositoryName}`);
+                    break;
+                } catch (error) {
+                    throw new LearnMoreError(`${Constants.SCAFFOLDING_PREREQUISITES} for C# module. ${error}`, "https://aka.ms/edge-csharp-module-prerequisites");
```
formulahendry: What if the error is not related to prerequisites not installed? e.g. network error or other error.

adashen: Good point. From the BI almost all the scaffolding part error is because of missing prerequisite. And the learn more page list the prerequisites and other information.

adashen: The ${error} will show the detail error message. Any suggestion about the wordings so the message be more clear?

formulahendry: `Please check whether the scaffolding prerequisites are installed`?

PR #463, opened 2019-06-14
Add telemetry to identify internal users

blackchoey:


src/common/utility.ts
```diff
@@ -622,6 +622,11 @@ export class Utility {
             validateFunc, Constants.moduleNameDft);
     }
 
+    public static isInternalUser(): boolean {
```
adashen: only telemetry related. So just put into telemetryClient class

src/common/telemetryClient.ts
```diff
@@ -4,14 +4,23 @@
...
 
 const packageJSON = vscode.extensions.getExtension(Constants.ExtensionId).packageJSON;
 const extensionVersion: string = packageJSON.version;
 const aiKey: string = packageJSON.aiKey;
+const isInternal: boolean = Utility.isInternalUser();
 
 export class TelemetryClient {
     public static sendEvent(eventName: string, properties?: { [key: string]: string; }): void {
+        if (properties) {
+            properties[Constants.isInternalPropertyName] = isInternal === true ? 'true' : 'false';
```
adashen: isInternal?'true':'false' ?

src/common/constants.ts
```diff
@@ -174,6 +174,8 @@ export class Constants {
     public static openSampleEvent = "openSample";
     public static openSampleUrlEvent = "openSampleUrl";
 
+    public static isInternalPropertyName = "IsInternal";
```
adashen: is the Uppercase of the first letter is a tradition in telemetry property?

blackchoey: I followed the VS and live share extension rule. Will update the case to align with our own style

src/common/telemetryClient.ts
blackchoey: use === to avoid implicit conversion in the comparison, though this should not happen in the isinternal scenario. Just want to make sure the comparison is robust if there's any changes on how to initialize isInternal value in the future.

Issue #467, opened 2019-06-21
the value of restartPolicy should be change from on-failed to on-failure


czgtest:
- VSCode Version: 1.35.1
- OS Version: windows/mac/ubuntu
- Extension Version:1.14.0

Steps to Reproduce:

1. create new solution for c# module
2. add  an "edgeModule" completion item as shown in screenshot
![image](https://user-images.githubusercontent.com/35018765/59912158-0ee5ad00-9448-11e9-85be-541e8c50cbbb.png)

3.Type enter, and the completion item will expand to a module configuration. 
You can use the tab key to visit each tabstop. 
For the status property, you can choose between “running” and “stopped”. For the restartPolicy property, you can choose among “always”, “never”, “on-failed”, and “on-unhealthy”. 
4. verify the “on-failed” is invalid
![image](https://user-images.githubusercontent.com/35018765/59911818-5a4b8b80-9447-11e9-9140-fefd825574d9.png)

Does this issue occur when all other extensions are disabled?: Yes/No

SLdragon:
From the schema, only "on-failure", “always”, “never”, and “on-unhealthy” are allowed

```
restartPolicy: {
enum: [
"never",
"on-failure",
"on-unhealthy",
"always"
]
}

```

czgtest:
fix work now, it will display "on-failure" for restartPolicy

PR #469, opened 2019-06-21
Fix intellisense for restartPolicy is not right

SLdragon:


SLdragon:
https://github.com/microsoft/vscode-azure-iot-edge/issues/467

PR #476, opened 2019-07-09
Guide user setup connection string from UI

SLdragon:


src/common/constants.ts
```diff
@@ -165,9 +165,12 @@ export class Constants {
     public static installFailedMsg = "'iotedgehubdev' tool installation has failed. Please install it manually for IoT Edge Simulator.";
     public static installManuallyMsg = "Please install 'iotedgehubdev' tool first for IoT Edge Simulator.";
     public static queryASAJobInfoFailedMsg = "The maximum retry count has been exceeded with empty response from the Stream Analytics.";
+    public static needSetupSimulatorMsg = "Please setup iotedgehubdev first before try to start simulator.";
```
adashen: before starting simulator

src/edge/simulator.ts
```diff
@@ -223,13 +225,44 @@ export class Simulator {
         });
     }
 
+    private async checkAndGuideUserToSetupIotedgehubdev(outputChannel: vscode.OutputChannel) {
+        if (await this.isValidateConfigSupported()) {
+            try {
+                await Executor.executeCMD(null, "iotedgehubdev", { shell: true }, "validateconfig");
+            } catch (error) {
```
adashen: will we check error code?

adashen: If we have error code, do we need is versionsupported ?

SLdragon: The error code is always -1 for current version

src/edge/simulator.ts
```diff
@@ -175,6 +175,7 @@ export class Simulator {
 
     public async startEdgeHubSingleModule(outputChannel: vscode.OutputChannel): Promise<void> {
         return await this.callWithInstallationCheck(outputChannel, async () => {
+            await this.checkAndGuideUserToSetupIotedgehubdev(outputChannel);
```
adashen: from the current flow, it will setup and continue to start?

src/common/constants.ts
SLdragon: Thanks, updated

src/edge/simulator.ts
SLdragon: updated

src/extension.ts
```diff
@@ -246,6 +246,8 @@ function initCommandAsync(context: vscode.ExtensionContext,
         let errorData: ErrorData | undefined;
         const properties: { [key: string]: string; } = {};
         properties.result = "Succeeded";
+        properties.fromCommandPalette = (!args || !args[0]).toString();
```
adashen: Can we make sure all the context menu triggered has args[0] not null?

SLdragon: Yes, args[0] for all commands not triggered from command palette is not null

PR #477, opened 2019-07-10
Read connection string from Toolkit, rename 'tempSensor' to 'Simulate…

formulahendry:
…dTemperatureSensor'

### DIFF
--- a/src/common/configuration.ts
+++ b/src/common/configuration.ts
@@ -1,5 +1,6 @@
 "use strict";
 import * as vscode from "vscode";
+import { Utility } from "./utility";
 
 export class Configuration {
     public static getConfiguration(): vscode.WorkspaceConfiguration {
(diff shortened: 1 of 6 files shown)


src/common/configuration.ts
```diff
@@ -18,8 +19,12 @@ export class Configuration {
...
 
-    public static getIotHubConnectionString(): string | undefined {
-        const toolkitConfig: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("azure-iot-toolkit");
-        return toolkitConfig ? toolkitConfig.get("iotHubConnectionString") : undefined;
+    public static async getIotHubConnectionString(): Promise<string | undefined> {
+        const toolkit = Utility.getToolkit();
+        try {
+            return toolkit.exports.azureIoTExplorer.getIotHubConnectionString();
+        } catch (error) {
+            return undefined;
```
adashen: how about have fallback to configuration? If user doesn't update the toolkit but have edge updated?

formulahendry: Updated

PR #484, opened 2019-07-25
Add hint: Node.js Module creation may take some time

formulahendry:


src/edge/edgeManager.ts
```diff
@@ -504,6 +504,7 @@ export class EdgeManager {
                 }
             case Constants.LANGUAGE_NODE:
                 try {
+                    outputChannel.appendLine("Node.js Module creation may take some time.");
```
adashen: better to have an estimation of the time?

formulahendry: Updated

Issue #504, opened 2019-10-10
Add Azure Stream Analytics Iot Edge Module in a Solution using Visual Studio Code


Ilheu:
Hi,
I'm trying to add an Azure Stream Analytics job in my Iot Edge Solution, using Visual Studio Code, but I keep receiving this error: "Parse Stream Analytics jobs info failed: The maximum retry count has been exceeded with empty response from the Stream Analytics."
![image](https://user-images.githubusercontent.com/8053606/66560217-d3820200-eb4e-11e9-9e75-8766aef515d9.png)

SLdragon:
Hi, @Ilheu , 
This error is because stream analytics jobs cannot response valid information after retry 3 times.
Maybe there is a network issue on your machine, or the configuration of ASA on the cloud is not correct. 
You can change your network environment or re-create an ASA and try again

Ilheu:
It is not a network problem, maybe an ASA problem! I don't know what to change in configurations ... Could it be caused by the blob storage that contains the ASA job ? 
The Diagnose says : "No health related issues found" ...
I created a new one, and I receive the same error ... 
Could it be caused by not having these fields filled ? 
![image](https://user-images.githubusercontent.com/8053606/66638883-b52f0b80-ec0d-11e9-8c82-7bb7cd009cdc.png)

SLdragon:
Hi, Ilheu, you can follow this tutorial to setup your own stream analytics jobs:
https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/iot-edge/tutorial-deploy-stream-analytics.md

Make sure to associate a storage account to your ASA:

![image](https://user-images.githubusercontent.com/5545529/66641298-090d1500-ec4d-11e9-8d5a-c50754ab9845.png)

Ilheu:
Hi SLdragon ! 
I already associated my ASA to a blob storage account ... But I still cannot add my ASA module in my Solution in Visual Studio Code ...

SLdragon:
Hi, Ilheu, can you show me your other configurations? Make sure your input, output and query are corrected.

Ilheu:
Hi SLdragon ! Thanks for the quick response !

I think my inputs, outputs and queries  are correct, because I can add the ASA by Portal ... 

![image](https://user-images.githubusercontent.com/8053606/66702924-c3ac1e80-ed04-11e9-9b71-d6126eee0bd8.png)

[...]

PR #505, opened 2019-10-23
Optimize ASA retry logic

SLdragon:
https://github.com/microsoft/vscode-azure-iot-edge/issues/504

image

### DIFF
--- a/src/container/streamAnalyticsManager.ts
+++ b/src/container/streamAnalyticsManager.ts
@@ -158,7 +158,7 @@ export class StreamAnalyticsManager {
                     } else {
                         throw new Error(Constants.queryASAJobInfoFailedMsg);
                     }
-                } else if (jobInfoResult.statusCode === 200){
+                } else if (jobInfoResult.statusCode === 200) {
                     const result = JSON.parse(jobInfoResult.body);
                     if (result.status === "Succeeded") {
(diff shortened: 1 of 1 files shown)


src/container/streamAnalyticsManager.ts
```diff
@@ -136,32 +138,42 @@ export class StreamAnalyticsManager {
...
+                    const result = JSON.parse(jobInfoResult.body);
+                    if (result.status === "Succeeded") {
+                        const info = JSON.parse(result.manifest);
+                        return info;
+                    } else {
+                        throw new Error(result.error.message);
+                    }
                 } else {
-                    throw new Error(result.error.message);
+                    throw new Error("http status code: " + jobInfoResult.statusCode);
```
adashen: is there error message in jobInfoResult?

src/container/streamAnalyticsManager.ts
```diff
@@ -136,32 +138,42 @@ export class StreamAnalyticsManager {
...
                 }
             }
         } catch (error) {
+            if (error instanceof UserCancelledError) {
```
adashen: How about change to if not instance of UserCancelledError, do formatting, So "throw error" common logic could be shared

SLdragon: Updated

PR #511, opened 2019-11-01
Add docker state detection

SLdragon:


src/extension.ts
```diff
@@ -5,7 +5,7 @@
 import * as tls from "tls";
 import * as vscode from "vscode";
 import { ConfigNotSetError } from "./common/ConfigNotSetError";
-import { Constants } from "./common/constants";
+import { Constants, DockerState } from "./common/constants";
```
adashen: DockerState needed in this file? It seems wrapped in Utiltiy

src/common/constants.ts
```diff
@@ -202,3 +202,11 @@ export enum ContainerState {
     NotRunning,
     NotFound,
 }
+
+export enum DockerState {
+    NotInstalled,
+    NotRunning,
+    PermissionDenied,
+    Runing,
```
adashen: Running

src/common/utility.ts
```diff
@@ -627,6 +627,61 @@ export class Utility {
...
 
+    public static async checkDockerState(outputChannel: vscode.OutputChannel) {
+        let state: DockerState;
+        let errorMsg: any;
+        try {
+            await Executor.executeCMD(outputChannel, "docker", { shell: true }, "version");
+            state = DockerState.Runing;
+        } catch (error) {
+            errorMsg = error;
+            if (os.platform() === "win32") {
```
adashen: cache the os.platform() value?

src/common/utility.ts
```diff
@@ -627,6 +627,61 @@ export class Utility {
             validateFunc, Constants.moduleNameDft);
     }
 
+    public static async checkDockerState(outputChannel: vscode.OutputChannel) {
+        let state: DockerState;
+        let errorMsg: any;
+        try {
+            await Executor.executeCMD(outputChannel, "docker", { shell: true }, "version");
+            state = DockerState.Runing;
+        } catch (error) {
```
adashen: Is error code same between "command not found" and "not running"?

adashen: Have we checked the error code of the following cases?

src/common/utility.ts
```diff
@@ -627,6 +627,61 @@ export class Utility {
...
+            }
+        }
+
+        if (errorMsg) {
+            const install: vscode.MessageItem = { title: "Install Docker" };
+            const troubleshooting: vscode.MessageItem = { title: "Troubleshooting" };
+            const cancel: vscode.MessageItem = { title: "Cancel" };
+            let input: vscode.MessageItem;
+            let helpUrl: string;
+            if (state === DockerState.NotInstalled) {
```
adashen: there is case state is not set. Shall we handle it, or state should assigned a default value?

adashen: switch may be more clear here

src/common/utility.ts
```diff
@@ -627,6 +627,61 @@ export class Utility {
...
+                    state = DockerState.NotRunning;
+                } else if ("permission denied") {
+                    state = DockerState.PermissionDenied;
+                } else {
+                    state = DockerState.Unknown;
+                }
+            }
+        }
+
+        if (errorMsg) {
```
adashen: A lot of constants here. How about move to constants?

src/common/utility.ts
SLdragon: Yes, they are both 1, so it is useless.

SLdragon: For linux , if docker is not installed, it would 127. And I will update to handle this

src/common/utility.ts
```diff
@@ -632,48 +632,56 @@ export class Utility {
...
-                } else if (error.message.indexOf("This error may also indicate that the docker daemon is not running") > -1) {
+                } else if (error.message.indexOf(Constants.dockerNotRunningErrorMsgPatternOnWindows) > -1) {
                     state = DockerState.NotRunning;
                 } else {
                     state = DockerState.Unknown;
                 }
-            } else if (os.platform() === "linux" || os.platform() === "darwin") {
-                if (error.message.indexOf("command not found") > -1 || error.message.match(/Command '.*?' not found/)) {
+            } else if (platform === "linux" || platform === "darwin") {
+                if (error.message.indexOf(Constants.commandNotFoundErrorMsgPatternOnLinux) > -1 || error.message.match(/Command '.*?' not found/) || error.errorCode === 127) {
```
adashen: put error code checking first?

SLdragon: OK

PR #514, opened 2019-11-04
Use standalone simulator for VSCode Edge extension

SLdragon:


package.json
```diff
@@ -291,6 +291,10 @@
         "azure-iot-edge.executor.env": {
           "type": "object",
           "description": "Environment variables for the command executor and Azure IoT Edge terminals"
+        },
+        "azure-iot-edge.simulator.version": {
+          "type": "string",
+          "description": "Specify the Azure IoT EdgeHub Dev Tool version (Only take effect on Windows for current version). Example: 0.12.0"
```
adashen: "Azure IoT EdgeHub Dev Tool" seems to be a new naming. User will be confused.

src/common/constants.ts
```diff
@@ -161,8 +161,7 @@ export class Constants {
     public static updateSimulatorMsg = "Update your 'iotedgehubdev' tool to the latest for the best experience for IoT Edge Simulator.";
     public static failedInstallSimulator = "Failed to install 'iotedgehubdev' tool because of error:";
     public static outputNoSimulatorMsg = "Cannot execute command since 'iotedgehubdev' is not installed. Please install it first for IoT Edge Simulator.";
-    public static pipNotFoundMsg = "'pip' not found. Cannot install 'iotedgehubdev' tool. Please install it manually for IoT Edge Simulator.";
-    public static installFailedMsg = "'iotedgehubdev' tool installation has failed. Please install it manually for IoT Edge Simulator.";
+    public static installStandaloneFailedMsg = "Failed to install 'iotedgehubdev' tool, please check the output channel for detailed error message.";
```
adashen: specify the output channel name?

src/edge/simulator.ts
```diff
@@ -23,61 +24,21 @@ enum InstallReturn {
...
-        if (await Simulator.simulatorInstalled()) {
-            return InstallReturn.Success;
-        } else {
-            TelemetryClient.sendEvent(`${telemetryName}.install`);
-            const installRes = await Simulator.installSimulatorWithPip(true, Constants.needSimulatorInstalledMsg, outputChannel);
-            TelemetryClient.sendEvent(`${telemetryName}.install.${InstallReturn[installRes]}`);
-            return installRes;
-        }
-    }
+    private static latestReleaseInfoUrl = "https://api.github.com/repos/Azure/iotedgehubdev/releases/latest";
```
adashen: how about create a short link?

src/edge/simulator.ts
```diff
@@ -91,85 +52,94 @@ export class Simulator {
...
-            return true;
+            const simulatorType = await this.getSimulatorType();
+            if (simulatorType === SimulatorType.NotInstalled) {
+                message =  Constants.needSimulatorInstalledMsg;
+                type = "install";
+            } else {
+                const version: string | null = await this.getCurrentSimulatorVersion();
+                if (version && semver.valid(version)) {
+                    const latestVersion: string | undefined = await this.getLatestSimulatorVersion();
+                    if (latestVersion && !semver.eq(latestVersion, version)) {
```
adashen: why not gt?

src/edge/simulator.ts
```diff
@@ -91,85 +52,94 @@ export class Simulator {
...
+
+                if (simulatorType === SimulatorType.Pip) {
+                    type = "upgradePipPackage";
+                } else {
+                    type = "upgradeStandalone";
+                }
+            }
         } catch (error) {
-            return false;
+            message =  Constants.needSimulatorInstalledMsg;
```
adashen: when will error be thrown?

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+        const output: string = await Executor.executeCMD(undefined, this.getSimulatorExecutorPath(true), { shell: true }, "--version");
+        const version: string | null = Simulator.extractVersion(output);
+        return version;
+    }
+
+    private async simulatorInstalled(): Promise<boolean> {
+        return await Simulator.checkCmdExist(this.getSimulatorExecutorPath(true));
+    }
+
+    private getSimulatorExecutorPath(forceUseCmd: boolean = false): string {
```
adashen: the return value of this method better to be cached in the class instance. We call this method everyplace to call iotedgehub related stuff. Besides, I think check the version on Activate is more cleaner.

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+        await fse.move(path.join(Simulator.WindowsStandaloneSimulatorFolder, "iotedgehubdev"), path.join(Simulator.WindowsStandaloneSimulatorFolder, version));
+    }
+
+    private async autoInstallSimulator(outputChannel: vscode.OutputChannel = null): Promise<InstallReturn> {
+        // auto install only supported on windows. For linux/macOS ask user install manually.
+        if (os.platform() !== "win32") {
+            return InstallReturn.NotSupported;
+        }
+
+        if (!this.isInstalling) {
```
adashen: what will return if isInstalling?

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+                executorPath = `"${simulatorPath}"`;
+            }
+        }
+        return executorPath;
+    }
+
+    private async downloadStandaloneSimulatorWithProgress() {
+        await vscode.window.withProgress({
+            location: vscode.ProgressLocation.Notification,
+            title: "Downloading and installing Azure IoT EdgeHub Dev Tool (iotedgehubdev)...",
```
adashen: const

src/edge/simulator.ts
adashen: If user configured the version, need to check version?

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+        });
+    }
+
+    private async downloadStandaloneSimulator() {
+        const userConfiguredVersion: string = this.getUserConfiguredVersion();
+        let version: string;
+        let binariesZipUrl: string;
+        if (userConfiguredVersion) {
+            version = userConfiguredVersion;
+            binariesZipUrl = `https://github.com/Azure/iotedgehubdev/releases/download/${version}/iotedgehubdev-${version}-win32-ia32.zip`;
```
adashen: create a short link for the github release url

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+        }
+
+        await new Promise((resolve, reject) => {
+            const req = request(binariesZipUrl);
+            req.on("response",  (res) => {
+                if (res.statusCode === 200) {
+                    req.pipe(unzipper.Extract({ path: Simulator.WindowsStandaloneSimulatorFolder }))
+                    .on("close", () => resolve()).on("error", (e) => reject(e));
+                } else if (res.statusCode === 404) {
+                    reject(new Error("Cannot download simulator, please check that the simulator version has been configured correctly."));
```
adashen: Also give a version sample?

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+                } else {
+                    reject(new Error("Cannot download simulator with status code: " + res.statusCode));
+                }
+            });
+        });
+
+        try {
+            const originalVersion: string = this.context.globalState.get(Simulator.SimulatorVersionKey);
+            if (originalVersion) {
+                fse.removeSync(path.join(Simulator.WindowsStandaloneSimulatorFolder, originalVersion));
```
adashen: why we use sync here?

src/edge/simulator.ts
```diff
@@ -226,10 +196,156 @@ export class Simulator {
...
+            }
+        } catch (err) {
+            // ignore
+        }
+
+        this.context.globalState.update(Simulator.SimulatorVersionKey, version);
+        await fse.move(path.join(Simulator.WindowsStandaloneSimulatorFolder, "iotedgehubdev"), path.join(Simulator.WindowsStandaloneSimulatorFolder, version));
```
adashen: instead of move, how about just download and extract to this folder?

src/edge/simulator.ts
SLdragon: https://aka.ms/iotedgehubdev-latest-release

src/edge/simulator.ts
SLdragon: Add a new status: InstallReturn.IsInstalling

src/edge/simulator.ts
SLdragon: Use a static  variable for root url :     
```
private static iotedgehubdevDownloadRootUrl = "https://github.com/Azure/iotedgehubdev/releases/download";
```

src/edge/simulator.ts
SLdragon: because zip root folder is 'iotedgehubdev', and we need to rename it when it extract completely.

src/edge/simulator.ts
SLdragon: Checked and found out that we do not need to handle error

src/edge/simulator.ts
SLdragon: Update this logic

src/common/constants.ts
```diff
@@ -161,7 +161,9 @@ export class Constants {
     public static updateSimulatorMsg = "Update your 'iotedgehubdev' tool to the latest for the best experience for IoT Edge Simulator.";
     public static failedInstallSimulator = "Failed to install 'iotedgehubdev' tool because of error:";
     public static outputNoSimulatorMsg = "Cannot execute command since 'iotedgehubdev' is not installed. Please install it first for IoT Edge Simulator.";
-    public static installStandaloneFailedMsg = "Failed to install 'iotedgehubdev' tool, please check the output channel for detailed error message.";
+    public static outputSimulatorIsInstallingMsg = "'iotedgehubdev' is installing now, please wait for installing to complete.";
```
adashen: wait for installation

src/edge/simulator.ts
```diff
@@ -106,9 +104,6 @@ export class Simulator {
                     type = "upgradeStandalone";
                 }
             }
-        } catch (error) {
-            message =  Constants.needSimulatorInstalledMsg;
-            type = "install";
         }
 
         if (message) {
```
adashen: Do we need to check this now? It seems if come here there should be message, else it already return?

SLdragon: Yes, we need to check it, because if user has already installed the latest version, we do not need to take any other actions.

SLdragon: another way is using return: 
image

src/common/constants.ts
SLdragon: How about this? 
'iotedgehubdev' is being installed now, and please wait for the installation.

adashen: ok

src/edge/simulator.ts
```diff
@@ -91,85 +56,85 @@ export class Simulator {
...
+                    message = Constants.updateSimulatorMsg;
+                }
 
-    private static async checkCmdExist(cmd: string): Promise<boolean> {
-        try {
-            await Executor.executeCMD(null, cmd, { shell: true }, "--version");
-            return true;
-        } catch (error) {
-            return false;
+                if (simulatorType === SimulatorType.Pip) {
```
adashen: so the logic is inconsistent here. If no message type should also not assigned.

SLdragon: OK, I think we can use return to avoid unnecessary code

src/edge/simulator.ts
```diff
@@ -226,10 +191,168 @@ export class Simulator {
...
+            }
+            return version;
+        } catch (error) {
+            return undefined;
+        }
+    }
+
+    private getUserConfiguredVersion(): string {
+        const version: string = Configuration.getConfigurationProperty("simulator.version");
+        if (version && version.match(/^\d+\.\d+\.\d+$/)) {
```
adashen: only allowed number in version?

adashen: for test package, we may have other chars

SLdragon: oh, yes, I will use loose check to allow rc release

src/edge/simulator.ts
adashen: Seems already return, so no need to check?

src/edge/simulator.ts
```diff
@@ -226,10 +194,148 @@ export class Simulator {
...
+                if (!forceUseCmd) {
+                    executorPath = `"${Utility.adjustFilePath(this.standaloneSimulatorInstalledPath)}"`;
+                    if (Utility.isUsingPowershell()) {
+                        executorPath = `& ${executorPath}`;
+                    }
+                } else {
+                    executorPath = `"${this.standaloneSimulatorInstalledPath}"`;
+                }
+            } else {
+                return "";
```
adashen: return null or undefined to be more clear.

src/edge/simulator.ts
```diff
@@ -226,10 +194,148 @@ export class Simulator {
...
+            const infoJson = JSON.parse(infoString);
+            this.latestStandaloneSimulatorInfoJson = infoJson;
+            return infoJson;
+        } else {
+            return this.latestStandaloneSimulatorInfoJson;
+        }
+    }
+
+    private async getSimulatorType(): Promise<SimulatorType> {
+        if (await this.simulatorInstalled()) {
```
adashen: seems simulatorInstalled  method should return whether standalone or pip, instead of re-check the platform here.

src/edge/simulator.ts
```diff
@@ -270,22 +376,31 @@ export class Simulator {
...
         switch (installReturn) {
             case InstallReturn.Success:
                 return await callback();
-            case InstallReturn.NoPip:
-                outputChannel.appendLine(Constants.outputNoSimulatorMsg);
-                throw new LearnMoreError(Constants.pipNotFoundMsg, Simulator.learnMoreUrl);
             case InstallReturn.Failed:
-                outputChannel.appendLine(Constants.outputNoSimulatorMsg);
-                throw new LearnMoreError(Constants.installFailedMsg, Simulator.learnMoreUrl);
+                await vscode.window.showErrorMessage(Constants.installStandaloneSimulatorFailedMsg);
```
adashen: why remove the output channel message?

adashen: No learn more message?

src/edge/simulator.ts
SLdragon: OK

src/edge/simulator.ts
SLdragon: Yes, learn more page is to help user to install pip package, I think we do not need it here.

src/edge/simulator.ts
SLdragon: updated

src/edge/simulator.ts
SLdragon: We also do not use output chanel here, because in this part, we cannot get deatailed error message. We already use output channel when error occured in "autoInstallSimulator" function

src/edge/simulator.ts
```diff
@@ -226,36 +228,34 @@ export class Simulator {
...
     }
 
     private async simulatorInstalled(): Promise<SimulatorType> {
-        const exist = await Simulator.checkCmdExist(this.getSimulatorExecutorPath(true));
-        if (exist) {
-            return Simulator.currentPlatform === "win32" ? SimulatorType.Standalone : SimulatorType.Pip;
-        } else {
-            return SimulatorType.NotInstalled;
+        if (this.standaloneSimulatorExecutablePath) {
+            const exist = await Simulator.checkCmdExist(this.getAdjustedSimulatorExecutorPath(true));
```
adashen: check the path before calling checkCmdExist?

src/edge/simulator.ts
```diff
@@ -91,85 +56,90 @@ export class Simulator {
         return null;
     }
 
-    private static async getLatestSimulatorVersion(): Promise<string | undefined> {
+    private static async checkCmdExist(cmd: string): Promise<boolean> {
```
adashen: add a string check here so if string is undefined/null just return false without exception.

src/edge/simulator.ts
```diff
@@ -91,85 +56,90 @@ export class Simulator {
...
 
-            let input: vscode.MessageItem | undefined;
-            if (force) {
-                input = await vscode.window.showWarningMessage(message, { modal: true }, ...items);
-            } else {
-                input = await vscode.window.showWarningMessage(message, ...items);
-            }
+    private isInstalling: boolean = false;
+    private latestStandaloneSimulatorInfoJson: any;
+    private standaloneSimulatorExecutablePath: string;
```
adashen: standalone may be confusing. "iotedgehubdev" is not standalone. Rename to simulatorExecutablePath?

src/edge/simulator.ts
```diff
@@ -91,85 +56,90 @@ export class Simulator {
...
+                type = "upgradePipPackage";
+            } else {
+                type = "upgradeStandalone";
+            }
         }
-    }
 
-    constructor(private context: vscode.ExtensionContext) {
+        TelemetryClient.sendEvent(`${telemetryName}.${type}`);
+        try {
```
adashen: how about move try upper to the entry of the method?

src/edge/simulator.ts
SLdragon: OK, check it inside checkCmdExist

src/edge/simulator.ts
SLdragon: OK

src/edge/simulator.ts
SLdragon: OK

SLdragon: We can use output channel to show unexpected errors

src/edge/simulator.ts
SLdragon: OK

PR #521, opened 2019-12-09
Send detailed error message when installed standalone simulator failed

SLdragon:


src/common/utility.ts
```diff
@@ -632,7 +632,7 @@ export class Utility {
         let state: DockerState;
         let errorMsg: any;
         try {
-            await Executor.executeCMD(outputChannel, "docker", { shell: true }, "version");
+            await Executor.executeCMD(outputChannel, "docker2", { shell: true }, "version");
```
adashen: docker2?

src/common/InstallResult.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
adashen: this file is only related to simulator right? How about to edge folder?

src/edge/simulator.ts
```diff
@@ -281,9 +282,7 @@ export class Simulator {
             req.on("response",  (res) => {
                 if (res.statusCode === 200) {
                     req.pipe(unzipper.Extract({ path: Simulator.WindowsStandaloneSimulatorFolder }))
-                    .on("close", () => resolve()).on("error", (e) => reject(e));
-                } else if (res.statusCode === 404) {
```
adashen: 404 is not specially handled?

src/edge/simulator.ts
```diff
@@ -201,6 +194,14 @@ export class Simulator {
         });
     }
 
+    private sendTelemetry(eventName: string, errMsg: string = null) {
```
adashen: what is the benefit of wrapping this method? same way to call telemetryclient?

src/common/InstallResult.ts
SLdragon: OK

src/edge/simulator.ts
SLdragon: Because use cannot set the version or download url manually, , so we do not need to special handle the 404 case

PR #531, opened 2020-03-11
Add repository url check

SLdragon:
Some cases:
```
aa/aaaAdd:888
invalid reference format: repository name must be lowercase

aaa/ddd/ddcc/eeAaa/aaadd:888
invalid reference format: repository name must be lowercase

AAA.dd*d/aa/bb:latest
invalid reference format

localHost/aa/bb:latest
invalid reference format: repository name must be lowercase

aaa:*ddd
invalid reference format

aaa/dddd/ddd/aa.aa 
valid

aa:88/aa/bb
valid

AAA.ddd/aa/bb:latest
valid

localhost/aa/bb:latest 
valid
```

### DIFF
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -587,11 +587,11 @@ export class Utility {
         }
 
         if (hostName && /[^a-zA-Z0-9\.\-:]/.test(hostName)) {
-            return "Repository host name contains invalid characters";
+            return "Repository host name can only contain alphanumeric characters or .-:";
         }
 
         if (/[^a-z0-9\._\-\/]+/.test(repositoryName)) {
-            return "Repository name contains invalid characters";
+            return "Repository name can only contain lowercase letters, digits or ._-/";
         }
 
         if (tag) {
@@ -599,8 +599,8 @@ export class Utility {
                 return "The maximum length of tag is 128";
             }
 
-            if (/[^a-zA-Z0-9\._\-\/]+/.test(tag)) {
-                return "Tag contains invalid characters";
+            if (/[^a-zA-Z0-9\._\-]+/.test(tag)) {
+                return "Tag can only contain alphanumeric characters or ._-";
             }
         }
 
--- a/src/common/utility.ts
+++ b/src/common/utility.ts
@@ -587,11 +587,11 @@ export class Utility {
         }
 
         if (hostName && /[^a-zA-Z0-9\.\-:]/.test(hostName)) {
-            return "Repository host name invalid characters";
+            return "Repository host name contains invalid characters";
         }
 
         if (/[^a-z0-9\._\-\/]+/.test(repositoryName)) {
-            return "Repository name contain invalid characters";
+            return "Repository name contains invalid characters";
         }
 
         if (tag) {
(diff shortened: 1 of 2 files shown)


src/common/utility.ts
```diff
@@ -566,6 +566,47 @@ export class Utility {
...
+        const tag: string = match[3];
+
+        // if host name do not contain ":.", then it would be treated as repository name
+        if (hostName && !/[:.]/.test(hostName)) {
+            repositoryName = hostName  + "/" + repositoryName;
+            hostName = null;
+        }
+
+        if (hostName && /[^a-zA-Z0-9\.\-:]/.test(hostName)) {
+            return "Repository host name contains invalid characters";
```
adashen: instead of show invalid. how about show what is valid characters?

src/common/utility.ts
```diff
@@ -566,6 +566,47 @@ export class Utility {
...
+        }
+
+        if (/[^a-z0-9\._\-\/]+/.test(repositoryName)) {
+            return "Repository name contains invalid characters";
```
adashen: same as above

src/common/utility.ts
```diff
@@ -566,6 +566,47 @@ export class Utility {
...
+        }
+
+        if (tag) {
+            if (tag.length > 128) {
```
adashen: the length is just for tag or the whole? (reposity+tag)

src/common/utility.ts
```diff
@@ -566,6 +566,47 @@ export class Utility {
...
+                return "The maximum length of tag is 128";
+            }
+
+            if (/[^a-zA-Z0-9\._\-\/]+/.test(tag)) {
+                return "Tag contains invalid characters";
```
adashen: same as above

src/common/utility.ts
SLdragon: OK, thanks, it's more reasonable

src/common/utility.ts
SLdragon: just for tag

src/common/utility.ts
```diff
@@ -566,6 +566,47 @@ export class Utility {
         return path.join(solutionPath, Constants.vscodeFolder, Constants.vscodeSettingsFile);
     }
 
+    public static async validateRepositoryUrl(repositoryUrl: string): Promise<string | undefined> {
+        if (!repositoryUrl) {
+            return "Repository url could not be empty";
+        }
+
+        const match: RegExpMatchArray = repositoryUrl.match(/^(?:([^\/]+)\/)?([^:]+)(?::(.+))?$/);
```
adashen: where did we get the regex or write by yourself?

adashen: Can we just get the first part before "/" like what we did in getting address?

SLdragon: myself, this regex is to divide the repository url to three part: hostName, repository name and tag.
However, the validate function cannot cover all invalid cases such as aa/bb///ccdd

src/common/utility.ts
SLdragon: updated

src/common/utility.ts
SLdragon: updated

src/common/utility.ts
SLdragon: > Can we just get the first part before "/" like what we did in getting address?

Yes, the first part of this regex : (?:([^\/]+)\/)?  is to get first part before "/" just like we use indexOf("/")

PR #535, opened 2020-03-20
Add single module debug for python, add arm64v8 platform

SLdragon:


package.json
```diff
@@ -271,6 +271,7 @@
           "default": {
             "amd64": [],
             "arm32v7": [],
+            "arm64v8":[],
```
blackchoey: What will happen if the module doesn't have arm64v8 dockerfile?

SLdragon: just like windows-amd64, some module templates doesn't contain  windows-amd64 dockerfile, if user change platform to windows-amd64, the deployment.template would show error tips

package.json
```diff
@@ -436,6 +437,94 @@
...
+            "pathMappings": [
+              {
+                "localRoot": "${workspaceFolder}/modules/<module-folder>",
+                "remoteRoot": "<Absolute path of the program in the docker container>"
+              }
+            ],
+            "windows": {
+              "pathMappings": [
+                {
+                  "localRoot": "${workspaceFolder}\\modules\\<module-folder>",
```
adashen: The folder may not go to modules folder. Since in this scenario, we may expect user enter from normal python project. Please refer to the root file of other languages

SLdragon: Updated

PR #536, opened 2020-03-25
Read target framework from csproj

SLdragon:
[AB#1699397](https://mseng.visualstudio.com/VSIoT/_workitems/edit/1699397)

### DIFF
--- a/src/edge/edgeManager.ts
+++ b/src/edge/edgeManager.ts
@@ -328,7 +328,7 @@ export class EdgeManager {
                 mapObj.set(Constants.appFolder, "/app");
                 const csprojPath: string = path.join(slnPath, Constants.moduleFolder, moduleName, moduleName + Constants.csharpProjectFlieExtensionName);
                 const csprojStr: string = await fse.readFile(csprojPath, "utf-8");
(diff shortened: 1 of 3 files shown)


src/edge/edgeManager.ts
```diff
@@ -325,6 +326,10 @@ export class EdgeManager {
             case Constants.LANGUAGE_CSHARP:
                 launchFile = Constants.launchCSharp;
                 mapObj.set(Constants.appFolder, "/app");
+                const csprojPath: string = path.join(slnPath, Constants.moduleFolder, moduleName, moduleName + Constants.csharpProjectFlieExtensionName);
+                const csprojStr: string = await fse.readFile(csprojPath, "utf-8");
+                const targetFramework: string = csprojStr.match(/<TargetFramework>(.+?)<\/TargetFramework>/)[1];
```
adashen: Is it possible there is no targetFramework?

SLdragon: targetFramework must be existing, otherwise, it would be an invalid csproj file. 
And this file is come from template and controlled by ourselves, so we can safely consider it must be existing

adashen: need trim spaces?

SLdragon: updated

PR #540, opened 2020-04-02
Update logic to valid env in repository url

SLdragon:


src/common/utility.ts
```diff
@@ -586,20 +586,25 @@ export class Utility {
             hostName = null;
         }
 
-        if (hostName && /[^a-zA-Z0-9\.\-:\${}]/.test(hostName)) {
-            return "Repository host name can only contain alphanumeric characters or .-:, and ${} are also supported for environment variables";
+        if (hostName) {
+            const hostNameWithoutEnv = hostName.replace(/\${[^${} ]+?}/g, "");
```
adashen, line 590: env could without {}

SLdragon: Updated

PR #544, opened 2020-05-13
Update error telemetry

SLdragon:


src/edge/simulator.ts
```diff
@@ -114,7 +114,13 @@ export class Simulator {
             TelemetryClient.sendEvent(`${telemetryName}.${type}`);
 
             const installResult = await this.autoInstallSimulator(outputChannel);
-            TelemetryClient.sendEvent(`${telemetryName}.${type}.${InstallReturn[installResult.resultType]}`, { error: installResult.errMsg });
+
+            if (installResult.errMsg) {
+                TelemetryClient.sendErrorEvent(`${telemetryName}.${type}.${InstallReturn[installResult.resultType]}`, { error: installResult.errMsg });
+            } else {
+                TelemetryClient.sendEvent(`${telemetryName}.${type}.${InstallReturn[installResult.resultType]}`);
```
adashen, line 121: there is no property in this scenario? So nothing will be sent?

SLdragon: Just send an event with the name such as "simulatorUpdated.install.Success"

src/common/telemetryClient.ts
```diff
@@ -12,22 +12,31 @@ const aiKey: string = packageJSON.aiKey;
...
-                [Constants.isInternalPropertyName] : this._isInternal === true ? "true" : "false",
-            };
-        }
+        this.stampInternalProperty(properties);
         this._client.sendTelemetryEvent(eventName, properties);
     }
 
+    public static sendErrorEvent(eventName: string, properties?: { [key: string]: string; }): void {
+        this.stampInternalProperty(properties);
+        this._client.sendTelemetryErrorEvent(eventName, properties, null, ["error", "errorMessage"]);
```
adashen, line 21: Define a const list of error properties? And when using error properties, we could ref the consts.

PR #569, opened 2021-03-25
Allow user to select between Edge Runtime 1.0 or 1.1

marianan:
When there is an active workspace, user can select which version of Edge Runtime is used. Default is 1.0 for both Edge agent and Edge hub.

To switch from the default, open the VSCode palette and select "Select Azure IoT Edge Solution Default Platform". The list of option right now is 1.0 and 1.1. Once the new version is selected, the existing modules in the workspace will have their deployment.json files updated and any new modules will use the last set desired Edge Runtime version.

The desired Edge Runtime version is used by simulator as well, in both solution and single module modes.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,6 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
-                //"--extensionDevelopmentPath=C:\\Work\\AzureIoT\\IoTEdge\\IoTEdgeHubDev\\Modules\\EdgeSolutionTest01"
                 "--extensionDevelopmentPath=${workspaceFolder}"
             ],
             "outFiles": [
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,6 +8,7 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
+                //"--extensionDevelopmentPath=C:\\Work\\AzureIoT\\IoTEdge\\IoTEdgeHubDev\\Modules\\EdgeSolutionTest01"
                 "--extensionDevelopmentPath=${workspaceFolder}"
             ],
             "outFiles": [
--- a/src/common/version.ts
+++ b/src/common/version.ts
@@ -9,6 +9,10 @@ export class Versions {
         return verMap;
     }
 
+    public static getSupportedEdgeRuntimeVersions(): string[] {
+        return Versions.getValue(Constants.versionEdgeRuntime, []) as string[];
+    }
+
     public static installCSharpTemplate(): boolean {
         return Versions.getValue(Constants.installCSharpModule, true) as boolean;
     }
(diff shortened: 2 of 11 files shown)


src/edge/edgeManager.ts
```diff
@@ -372,6 +393,28 @@ export class EdgeManager {
         }
     }
 
+    private async updateRuntimeVersionInDeploymentTemplate() {
+        const pattern = `{${Constants.deploymentTsonPattern}}`;
```
mhshami01, line 397: 'J'?

src/edge/simulator.ts
```diff
@@ -156,8 +157,9 @@ export class Simulator {
         return await this.callWithInstallationCheck(outputChannel, async () => {
             await this.checkIoTedgehubdevConnectionString(outputChannel);
             const inputs = await this.inputInputNames();
+            const imgVersion = Versions.edgeHubVersion();
             await this.setModuleCred(outputChannel);
-            await Executor.runInTerminal(Simulator.adjustTerminalCommand(this.getAdjustedSimulatorExecutorPath() + ` start -i "${inputs}"`));
+            await Executor.runInTerminal(Simulator.adjustTerminalCommand(this.getAdjustedSimulatorExecutorPath() + ` start -img "${imgVersion}" -i "${inputs}"`));
```
mhshami01, line 162: -img => -er if using edge hub dev 0.14.6 or greater :-)

.vscode/launch.json
```diff
@@ -8,6 +8,7 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
+                //"--extensionDevelopmentPath=C:\\Work\\AzureIoT\\IoTEdge\\IoTEdgeHubDev\\Modules\\EdgeSolutionTest01"
```
mhshami01, line 11: I recommend deleting this line

mhshami01 (changes requested):

src/edge/edgeManager.ts
marianan: reused existing variable, will rename it, thanks for catching

---
In reply to: [602420877](https://github.com/microsoft/vscode-azure-iot-edge/pull/569#discussion_r602420877) [](ancestors = 602420877)

mhshami01 (approved):

mhshami01 (approved):

src/common/version.ts
```diff
@@ -49,15 +53,42 @@ export class Versions {
...
+        if (templateJson) {
+            const sysModuleImage =
+                templateJson.modulesContent.$edgeAgent["properties.desired"].systemModules[moduleName].settings.image;
+            templateJson.modulesContent.$edgeAgent["properties.desired"].systemModules[moduleName].settings.image =
+                Versions.getNewImageVersionJson(sysModuleImage, versionMap);
+        }
+    }
+
+    public static edgeHubVersion(): string {
+        return Versions.getDefaultEdgeRuntimeVersion();
```
YoDaMa, line 66: this resolves to 1.0 always, correct? Shouldn't it resolve to 1.1 or 1.0 depending on the user selected Edge runtime?

src/common/version.ts
```diff
@@ -49,15 +53,42 @@ export class Versions {
...
     }
 
-    private static edgeHubVersion(): string {
-        return Versions.getValue(Constants.versionEdgeHub, "1.0") as string;
+    private static getDefaultEdgeRuntimeVersion(): string {
+        return Versions.getValue(Constants.versionDefaultEdgeRuntime, "1.0") as string;
+    }
+
+    private static getNewImageVersionJson(input: string, versionMap: Map<string, string>): string {
```
YoDaMa, line 77: maybe you should define a type for the imageName, since it looks like you're expecting a certain form of string.

YoDaMa: something like:
```ts
type  fooType =`$${string}:${string}`
```

also based on the type definitions, `input` cannot be undefined, so `if (input) {' is both vague and redundant. If you're checking that it's not an empty string, maybe define it as `input !== ''`

src/common/version.ts
```diff
@@ -49,15 +53,42 @@ export class Versions {
         return Versions.getValue(Constants.versionTempSensor, "1.0") as string;
     }
 
+    public static updateSystemModuleImageVersion(templateJson: any, moduleName: string, versionMap: Map<string, string>) {
+        if (templateJson) {
```
YoDaMa, line 57: are you checking if templateJson is undefined? Because of JS falsyness being vague, better to be explicit here.

src/edge/edgeManager.ts
```diff
@@ -156,6 +156,27 @@ export class EdgeManager {
...
+        TelemetryClient.sendEvent(`${Constants.selectEdgeRuntimeVerEvent}`, {
+            template: edgeVersionPick,
+        });
+
+        await Configuration.setWorkspaceConfigurationProperty(Constants.versionDefaultEdgeRuntime, edgeVersionPick);
+        outputChannel.appendLine(`Default Azure IoT Edge Runtime is ${edgeVersionPick} now.`);
+
+        // If there is an active workspace, update the deployment templates
+        // with the desired runtime version
+        if (Utility.checkWorkspace()) {
```
YoDaMa, line 175: `Utility.checkWorkspace() !== undefined`?

YoDaMa (approved):
approved with nit comments

src/common/version.ts
marianan: when the user selects a new runtime version, it becomes the new default version, the workspace config is updated and getDefaultEdgeRuntimeVersion will return whatever the latest default version is selected. I call it default version because after it has been selected, all new modules will be generated using this version.

---
In reply to: [602606639](https://github.com/microsoft/vscode-azure-iot-edge/pull/569#discussion_r602606639) [](ancestors = 602606639)

mhshami01 (approved):

PR #586, opened 2021-08-05
Deployment schema validation supporting Edge Runtime 1.2

marianan:
This change creates a new deployment schema file to support intellisense and validation against the edgeHub and edgeAgent split schemas.

The default Edge Runtime for new modules is now 1.2 and can be adjusted to earlier version if needed.


src/edge/simulator.ts
```diff
@@ -31,9 +31,9 @@ enum SimulatorType {
 }
 
 export class Simulator {
-    private static iotedgehubdevVersionUrl: string = "https://pypi.org/pypi/iotedgehubdev/json";
+    private static iotedgehubdevVersionUrl: string = "https://test.pypi.org/pypi/iotedgehubdev/json";
     private static iotedgehubdevLockVersionKey = "IOTEDGEHUBDEV_VERSION";
-    private static iotedgehubdevDefaultVersion = "0.14.8";
+    private static iotedgehubdevDefaultVersion = "0.14.9rc0";
```
marianan, line 36: >9rc0 [](http://example.com/codeflow?start=55&length=4)

will update to 0.14.9 after testing is completed

assets/solution/deployment.template.schema.json
```diff
@@ -0,0 +1,26 @@
+{
```
marianan, line 1: will discuss posting this schema to the schema store instead of storing locally in the developer sources

mhshami01 (approved):

package.json
```diff
@@ -308,7 +308,7 @@
     "jsonValidation": [
       {
         "fileMatch": "deployment.template.json",
-        "url": "http://json.schemastore.org/azure-iot-edge-deployment-template-2.0"
+        "url": "http://json.schemastore.org/azure-iot-edge-deployment-template-4.0"
       },
       {
         "fileMatch": "deployment.*.template.json",
```
mhshami01, line 314: should this also use the 4.0 schema?

assets/solution/deployment.template.json
```diff
@@ -16,7 +16,7 @@
           "edgeAgent": {
```
mhshami01, line 16: Should we update the 'schema-template'?

mhshami01:
Do we need an 'upgrade' path? What do we do when a user opens a solution generated with a previous release of the extension?

marianan:
the user has an option to upgrade using the Update Edge Runtime version, then the deployment template files will be updated with proper versions.

---
In reply to: [895532482](https://github.com/microsoft/vscode-azure-iot-edge/pull/586#issuecomment-895532482) [](http://example.com/codeflow?ancestors=895532482,895532482)

mhshami01 (approved):

PR #591, opened 2021-09-21
Dev/hshami/new dev containers

mhshami01:
Initial CSharp dev-container definition files.


containers/CSharp/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,74 @@
...
+    // Add the IDs of extensions you want installed when the container is created.
+    "extensions": [
+        "ms-dotnettools.csharp",
+        "ms-vscode.azure-account",
+        "ms-azuretools.vscode-docker",
+        "ms-vscode.azure-account",
+        "vsciot-vscode.azure-iot-hub",
+        "vsciot-vscode.azure-iot-edge",
+        "vsciot-vscode.azure-iot-toolkit",
+        "vsciot-vscode.azure-iot-tools",
```
marianan, line 39: >iot-tools [](http://example.com/codeflow?start=29&length=9)

this extension already packages iot-hub and iot-edge so we probably don't need both

containers/CSharp/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,74 @@
...
+        "source=ostf-bashhistory,target=/commandhistory,type=volume",
+        // Use hosts docker socket
+        "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind",
+    ],
+
+    // Set *default* container specific settings.json values on container create.
+    "settings": {},
+
+    // Add the IDs of extensions you want installed when the container is created.
+    "extensions": [
```
marianan, line 31: >extensions [](http://example.com/codeflow?start=5&length=10)

these extensions are installed on the VSCode instance that is running inside the container, not the dev machine one, right?

containers/CSharp/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,74 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.194.0/containers/dotnet
+{
+    "name": "C# (.NET)",
```
marianan, line 4: >C# (.NET) [](http://example.com/codeflow?start=13&length=9)

This is not a general purpose C# dev container, so we should call it Azure IoT Edge C# or something similar

containers/CSharp/.devcontainer/Dockerfile
```diff
@@ -0,0 +1,45 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.194.0/containers/dotnet/.devcontainer/base.Dockerfile
+
+# [Choice] .NET version: 5.0, 3.1, 2.1
+ARG VARIANT="5.0"
+FROM mcr.microsoft.com/vscode/devcontainers/dotnet:0-${VARIANT}
+
+# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
+ARG NODE_VERSION="none"
```
marianan, line 8: >"none" [](http://example.com/codeflow?start=17&length=6)

is there a default version that we should offer or do we always want the user to pick one?

marianan, line 8: >NODE_VERSION [](http://example.com/codeflow?start=4&length=12)

actually do you know why node is needed for teh C# modules?

containers/CSharp/.devcontainer/devcontainer.json
mhshami01: yes - thank you!

containers/CSharp/.devcontainer/devcontainer.json
mhshami01: Yes - I tested this with a bare-bone VSCode installation to verify :-)

marianan (approved):
:shipit:

PR #594, opened 2021-09-24
Dev/hshami/new python dev container

mhshami01:
python dev-container definitions


containers/Python/.devcontainer/library-scripts/azcli-debian.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+#
```
marianan, line 6: this library is common across all decontainers, should we package it in a way so they can all reuse without duplicating the code?

PR #597, opened 2021-09-24
new c/c++ devcontainer

cindydeng1998:


containers/Cpp/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,43 @@
...
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"ms-vscode.azure-account",
+		"ms-vscode.cpptools",
+        "ms-azuretools.vscode-docker",
+        "ms-vscode.azure-account",
+        "vsciot-vscode.azure-iot-toolkit",
+        "vsciot-vscode.azure-iot-tools",
+		"vsciot-vscode.vscode-iot-device-cube"
```
marianan, line 32: we should remove azure-account, iot-toolkit and device-cube as these are dependencies that are brought through iot-tools

PR #599, opened 2021-09-30
Vikas0212git patch 1

vikas0212git:
Updated the vscode file for GH actions.


.github/workflows/vscode.yml
```diff
@@ -0,0 +1,70 @@
...
+         npm run tslint
+         npm run compile
+         node scripts/genAiKey.js
+         echo "y" | vsce package
+    - name: npm test
+      run: xvfb-run --auto-servernum --server-args="-screen 0, 1024x768x24" npm test --silent
+    - name: docker test
+      run: docker run -t --rm -v $PWD:/mnt:ro dkhamsing/awesome_bot --white-list "gitter.im/Microsoft/vscode-azure-iot-edge" --allow-dupe --allow-redirect --skip-save-results `ls *.md`
+
+    - name: GitHub Release
```
marianan, line 37: >name: GitHub Release [](http://example.com/codeflow?start=6&length=20)

please add a condition that this step executes only if the 2 testing steps before have succeeded

.github/workflows/vscode.yml
```diff
@@ -0,0 +1,70 @@
...
+      uses: xresloader/upload-to-github-release@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        file: "*.vsix"
+
+    - name: Publish to Marketplace
```
marianan, line 44: >Publish to Marketplace [](http://example.com/codeflow?start=12&length=22)

this step should also execute only if the tests have succeeded

PR #601, opened 2021-09-30
Vikas0212git patch 1

vikas0212git:
Updated the file to enable tagged build


.github/workflows/vscode.yml
```diff
@@ -0,0 +1,69 @@
+name: CI
+
+on:
```
marianan, line 3: >on [](http://example.com/codeflow?start=0&length=2)

this pipeline should also work on PRs and merges to main

vikas0212git: I have changed that can you please take a look.

marianan (approved):
:shipit:

PR #608, opened 2021-10-05
Updated vscode.yml to release only for Semantic versions and genAikey.js in scripts to read GITHUB_TAG.

vikas0212git:


marianan (approved):
:shipit:

PR #615, opened 2021-10-19
User/marine/packagesupgrade

mhshami01:


CHANGELOG.md
```diff
@@ -12,6 +16,7 @@
 ## 1.24.4 - 2021-10-01
 ### Changed
 * Maximum version number for edgeHub/properties.desired.schemaVersion is capped to 1.1
+>>>>>>> master
```
marianan, line 19: nit: pls remove

marianan:
>              sudo npm install -g vsce

Should this also be updated to "npm install -g vsce"?

---
Refers to: .github/workflows/vscode.yml:81 in f0d4dd1. [](commit_id = f0d4dd1b875ae579440b365cb8bc511fa9433d41, deletion_comment = False)

marianan:
>     private async loadAmlImageItems(workspace: Workspace, session: AzureSession): Promise<vscode.QuickPickItem[]> {

do we have existing test coverage for the updated implementation?

---
Refers to: src/container/amlManager.ts:82 in f0d4dd1. [](commit_id = f0d4dd1b875ae579440b365cb8bc511fa9433d41, deletion_comment = False)

mhshami01:
> > ```
> >          sudo npm install -g vsce
> > ```
> 
> Should this also be updated to "npm install -g vsce"?
> 
> Refers to: .github/workflows/vscode.yml:81 in [f0d4dd1](https://github.com/microsoft/vscode-azure-iot-edge/commit/f0d4dd1b875ae579440b365cb8bc511fa9433d41). [](commit_id = [f0d4dd1](https://github.com/microsoft/vscode-azure-iot-edge/commit/f0d4dd1b875ae579440b365cb8bc511fa9433d41), deletion_comment = False)

No - I removed sudo from the Linux leg because it was not working. Mac is fine.

package.json
```diff
@@ -535,40 +535,42 @@
...
-    "postinstall": "node ./node_modules/vscode/bin/install",
     "tslint": "tslint -t verbose src/**/*.ts",
-    "test": "node ./node_modules/vscode/bin/test",
+    "test": "tsc -p ./ && node ./out/test/runTests.js",
     "webpack": "webpack --mode development",
     "webpack-dev": "webpack --mode development --watch"
   },
   "devDependencies": {
+    "@azure/arm-resources": "^4.2.2",
+    "@azure/arm-resources-subscriptions": "^2.0.0-beta.1",
```
marianan, line 545: >beta [](http://example.com/codeflow?start=50&length=4)

is there a stable version we could use? I worry of taking dependencies on a beta version.

test/utility.test.ts
```diff
@@ -179,7 +181,7 @@ suite("utility tests", () => {
         windows: null,
         test: ["test"],
       });
-      return stubMap;
+      return (stubMap as unknown) as vscode.WorkspaceConfiguration;
```
marianan, line 184: >unknown [](http://example.com/codeflow?start=25&length=7)

If the workspaceconfiguration isn't a map should we instead have a conversion utility than cast to unknown?

package.json
mhshami01: There are exactly two releases of this module, 1.0.0-beta.1 or 2.0.0-beta.1 [https://github.com/Azure/azure-sdk-for-js/releases?q=arm-resources-subscriptions&expanded=true](url)

test/utility.test.ts
mhshami01: This is test code and as such if we don't get a map returned, we have a bug and the test will fail. Leaving it as is will show the true nature of the bug. Furthermore, we are not changing the code or the logic here. The double cast was necessitated by stricter semantic rule enforcement in the new version.

marianan (approved):
:shipit:

PR #622, opened 2021-12-14
Dev/hshami/request to axios

mhshami01:
Replace deprecated Request package with the latest Axios.
Prevent changing of the name of AbortSignal constructor during webpack optimization to fix a GET issue.


webpack.config.js
```diff
@@ -74,4 +76,29 @@ const config = {
     ]
 }
 
-module.exports = config;
+// in production mode, webpack decorates class constructor names.
+// changing constructor names causes an issue for the following logic:
+//    if type.constructor.name !== 'AbortSignal'...
+// this issue is present in the package node-fetch - version "2.6.0", which is
+// used by the package @azure/ms-rest-js - version "2.6.0".
```
marianan, line 83: >@azure/ms-rest-js - version "2.6.0" [](http://example.com/codeflow?start=23&length=35)

is this a workaround only for ms-rest-js 2.6.0 version? Is there a newer version of ms-rest-js that we can snap to to avoid this workaround?

marianan (approved):
:shipit:

webpack.config.js
mhshami01: There is not - 2.6.0 is the latest :(

PR #626, opened 2022-01-27
Fixes error when creating a new solution with external modules

mhshami01:
This fixes error when creating a new solution with an external module as it's first module.


marianan (approved):
:shipit:

PR #632, opened 2022-04-01
Vikas/depalerts

vikas0212git:
Resolved the minimist by listing as a dev dependency rather than a sub dependency to tslint>mkdirp>minimist.


marianan (approved):
:shipit:

